PR #5, opened 2018-02-22
Viite 1126 current and history floating at same time

hestrompa:


hestrompa:
Running the tests before merging

PR #7, opened 2018-02-23
VIITE-1142 -Typo and wrong number in ELY-border validation message

pedrosag:
Created an additional case object with the purpose of manipulating the message of the Validation Error.


pedrosag:
Included in this PR is now the solution to the Viite-1163 task. As the main solution ALSO works on it.

hestrompa:
Merging this next

PR #18, opened 2018-03-01
VIITE-1117_change_road_styling_of_municipal_roads_based_to_road_addres.road_type_instead_of_vvh

pedrosag:
Backend:   
    .Added a additional call to the end of each fetch option in order to determine which sequence of addresses (by number and part number) get to have a black line underneath it.
    .Added said new parameter to all address objects, that go out to the UI.

UI:
    .Reworked the way the LinkProperty Styler handles the previously mentioned adminClassLine in order to depend on the newly created parameter from the backend.

### DIFF
--- a/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
+++ b/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
@@ -22,6 +22,7 @@ import org.scalatra.json.JacksonJsonSupport
 import org.scalatra.{NotFound, _}
 import org.slf4j.LoggerFactory
 
+import scala.reflect.macros.blackbox
 import scala.util.parsing.json._
 import scala.util.{Left, Right}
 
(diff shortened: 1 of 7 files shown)


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -765,7 +766,8 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
       "endMValue" -> roadAddressLink.endMValue,
       "sideCode" -> roadAddressLink.sideCode.value,
       "linkType" -> roadAddressLink.linkType.value,
-      "roadLinkSource" -> roadAddressLink.roadLinkSource.value
+      "roadLinkSource" -> roadAddressLink.roadLinkSource.value,
+      "blackUnderline" -> roadAddressLink.blackUnderline
```
oappicgi: wouldn't it be sufficient to have "blackUnderline" -> roadAddressLink.roadType.value==3  <ofc 3 being the variable but you get the idea>?

oappicgi: was it because last line in the story?  I think it was written  just so it would be easier to implement. Thus "future story remark".

pedrosag: No it would not, in the bottom of the user story it states that "For now, the styling should use stroke only in the situation where 50% or over of the total length of the link has ROAD_TYPE value 3.
" so with that in mind I made that precise check on the backend and condensed the result to a single boolean value (because evaluating if the total geometry length of the parts that have 3 as the road type is >= 50% of the total geometry of the road in the front end would just be a hog of resources) therefore the parameter is passed so that the styler can pick it up with ease.

PR #19, opened 2018-03-02
Viite refactoring to roadlinkservice

hestrompa:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
```diff
@@ -106,116 +111,78 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
...
+  /**
+    *
+    * @param bounds
+    * @param frozenTimeVVHAPIServiceEnabled
+    * @return Road links and change data from frozen API if frozenTimeVVHAPIServiceEnabled is true. Otherwise returns from normal VVH interface
+    */
+  def getRoadLinksAndChangesFromVVHWithFrozenAPI(bounds: BoundingRectangle, frozenTimeVVHAPIServiceEnabled: Boolean = false): (Seq[RoadLink], Seq[ChangeInfo]) = {
     if (frozenTimeVVHAPIServiceEnabled) {
       val (changes, links) =
-        Await.result(Future(Seq.empty[ChangeInfo]).zip(vvhClient.frozenTimeRoadLinkData.fetchByMunicipalitiesAndBoundsF(bounds, municipalities)), atMost = Duration.Inf)
```
oappicgi: think having municipality would make it faster for VVH to search links, is there a reason why those have been removed?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
```diff
@@ -106,116 +111,78 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
...
+        Await.result(Future(Seq.empty[ChangeInfo]).zip(vvhClient.frozenTimeRoadLinkData.fetchByMunicipalitiesAndBoundsF(bounds, Set[Int]())), atMost = Duration.Inf)
```
oappicgi: think having municipality would make it faster for VVH to search links, is there a reason why those have been removed?

hestrompa: Because the only time that we are calling the method is:
roadLinkService.getRoadLinksAndChangesFromVVHWithFrozenTimeAPI(boundingRectangle,Set(),frozenTimeVVHAPIServiceEnabled)._1

and we pass an empty set as the set of municipalities... So  just removed it, since it is never used as a parameter

PR #23, opened 2018-03-07
VIITE-1183 - Addressing discontinuity validation false positives

pedrosag:
Added another check after the main finding of invalid road addresses that test if they are part of the same road address, are continuous, have the same track code and are adjacent if they are then we discard them.


pedrosag:
@hestrompa I won't happen again, never again. The tests are mounted the code is solid. It is now a time to preform beyond one's limit.

PR #37, opened 2018-03-16
VIITE-1213 Fix in new road parts re-reservation

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1005,14 +1008,14 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
 
-    def checkAndMakeReservation(projectLink: ProjectLink) = {
+    def checkAndMakeReservation(projectLinks: Seq[ProjectLink]) = {
       val project = getProjectWithReservationChecks(projectId, newRoadNumber, newRoadPartNumber)
-      if (!project.isReserved(newRoadNumber, newRoadPartNumber)){
-        if(isNewAndSameLink(projectLink)){
-          val reservedPart = ProjectDAO.fetchReservedRoadPart(projectLink.roadNumber, projectLink.roadPartNumber)
+      if (!project.isReserved(newRoadNumber, newRoadPartNumber)) {
+        if (isCompletelyNewPart(projectLinks)) {
+          val reservedPart = ProjectDAO.fetchReservedRoadPart(projectLinks.head.roadNumber, projectLinks.head.roadPartNumber)
```
skosonen: Are we sure that all the ProjecLinks in projectLinks sequence have always the same road number and road part number here?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -990,8 +990,11 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
                          discontinuity: Int = Discontinuity.Continuous.value, ely: Option[Long] = None,
                          reversed: Boolean = false): Option[String] = {
 
-    def isNewAndSameLink(projectLink: ProjectLink): Boolean = {
-      linkIds.contains(projectLink.linkId) && (projectLink.status == linkStatus) && (linkStatus == New)
+    def isCompletelyNewPart(projectLinks: Seq[ProjectLink]): Boolean = {
```
skosonen: projectLinks parameter is never read and should be removed.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -998,14 +1002,14 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
 
-    def checkAndMakeReservation(projectLink: ProjectLink) = {
+    def checkAndMakeReservation(projectLinks: Seq[ProjectLink]) = {
       val project = getProjectWithReservationChecks(projectId, newRoadNumber, newRoadPartNumber)
-      if (!project.isReserved(newRoadNumber, newRoadPartNumber)){
-        if(isNewAndSameLink(projectLink)){
-          val reservedPart = ProjectDAO.fetchReservedRoadPart(projectLink.roadNumber, projectLink.roadPartNumber)
+      if (!project.isReserved(newRoadNumber, newRoadPartNumber)) {
+        if (isCompletelyNewPart()) {
+          val reservedPart = ProjectDAO.fetchReservedRoadPart(projectLinks.head.roadNumber, projectLinks.head.roadPartNumber)
```
skosonen: Are we sure that all the ProjecLinks in projectLinks sequence have always the same road number and road part number here?

Campanha: I think yes, at least every update must send always same road/part/track etc for any selected linkId's.

Campanha: Multiple selection w/ Shift key is allowed in same group (road, part, and not sure if also for track code)

skosonen: Maybe this should be commented in the code that it doesn't look like a possible bug?

Campanha: hm which specific line? At my point of view we could only have problems with this in case the UI change its logic in Shift-multiple selection by starting to send different roads parts whithin selected linkIds

skosonen: Line 1009, where we just take the head of projectLinks. When looking at this method without knowing how this method is used, this looks like an error. If there is even a small change that later this would be used with multiple road parts, then there could be also a check that project links have all the same road number and road part number. And if there are multiple different ones, then raise an exception and/or log an error.

Campanha: yeah ure right! I will add a comment on that saying how the method should be used and also add the validation check for all projectLinks to have all same part due to the current logic

PR #39, opened 2018-03-19
VIITE-82 Checks that common history id is handled correctly

skosonen:


skosonen:
Pull request "Viite common history project links implementation" should be merged first since the changes made in that branch are also included in this branch. Makes it easier to review the changes made in VIITE-82.

PR #42, opened 2018-03-20
Viite 1212 fuse road address history

plourenco90:


plourenco90:
You right, its now able to fuse all road addresses

PR #44, opened 2018-03-21
VIITE-1218 apply changes filter

plourenco90:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -318,8 +312,17 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
     sanityCheckedTypeOneTwo ++ nonCheckedChangeTypes
   }
 
-  def applyChanges(roadLinks: Seq[RoadLink], changes: Seq[ChangeInfo], addresses:
-  Map[(Long, Long), LinkRoadAddressHistory]): Seq[LinkRoadAddressHistory] = {
+  def applyChanges(roadLinks: Seq[RoadLink], allChanges: Seq[ChangeInfo], roadAddresses: Seq[RoadAddress]): Seq[LinkRoadAddressHistory] = {
+
+    def filterChanges: Seq[ChangeInfo] = {
+      val groupedAddresses = roadAddresses.groupBy(_.linkId)
+      val timestamps = groupedAddresses.mapValues(_.map(_.adjustedTimestamp).min)
```
pedrosag: Quick question, if the groupedAddresses is empty wouldn't the _.map(_.adjustedTimestamp).min) blow up in the .min part?

leirasf: It's true, but since you are inside a `mapValues` of a `groupby` you will have always at least one.

PR #49, opened 2018-03-26
Viite-1226

pedrosag:
Added the ability to fetch RoadLink data from Suravage interfaces.


hestrompa:
Needs attention because has failing tests

pedrosag:
Tests are passing , check build: 637 on jenkins.

PR #56, opened 2018-04-03
Viite 1228 and VIITE-1234 wrong loop directions and allow discontinuity drop down on numbering

leirasf:


digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
```diff
@@ -289,6 +289,38 @@ object GeometryUtils {
     (180 + (rad * (180 / Math.PI))).asInstanceOf[Int]
   }
 
+  def splitFromPoint(geometry: Seq[Point], point: Point, epsilon: Double): Option[Seq[Point]] = {
```
skosonen: Maybe epsilon could have default value here:
`  def splitFromPoint(geometry: Seq[Point], point: Point, epsilon: Double = DefaultEpsilon): Option[Seq[Point]] = {
`

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
```diff
@@ -289,6 +289,38 @@ object GeometryUtils {
...
+
+  def middlePoint(geometries: Seq[Seq[Point]]): Point = {
+    val minX = geometries.map(_.map(_.x).min).min
+    val maxX = geometries.map(_.map(_.x).max).max
+    val minY = geometries.map(_.map(_.y).min).min
+    val maxY = geometries.map(_.map(_.y).max).max
+    Point(minX + ((maxX - minX) / 2), minY + ((maxY - minY) / 2))
+  }
+
+  def connectionPoint(geometries: Seq[Seq[Point]], epsilon: Double): Option[Point] = {
```
skosonen: Epsilon could have default value here:
`  def splitFromPoint(geometry: Seq[Point], point: Point, epsilon: Double = DefaultEpsilon): Option[Seq[Point]] = {
`

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/Point.scala
```diff
@@ -85,6 +85,10 @@ case class Vector3d(x: Double, y: Double, z: Double) {
   def to2D(): Vector3d = {
     Vector3d(x, y, 0.0)
   }
+
+  def angle(that: Vector3d): Double = {
```
skosonen: This function should be renamed to something else, since this is not calculating the angle between two 3d vectors as one would expect. Later someone might use this angle function and expect that also the z-component affects the calculation and that the possible return values would be between [0, π]. Now this implementation returns also negative angles and doesn't take z-component in account.

On the other hand if we fix this "angle" function implementation, the "pickRightMost"-function would probably be broken if it relies on this that the function returns also negative values.

skosonen: Created new method with the functionality that the pickRightMost function needs and changed this angle-method to do what I would expect it to do.

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/Point.scala
```diff
@@ -85,6 +85,14 @@ case class Vector3d(x: Double, y: Double, z: Double) {
...
+    2.0d * Math.atan((this - that).length / (this + that).length)
+  }
+
+  def angleXYWithNegativeValues(that: Vector3d): Double = {
```
skosonen: This method should have a better name.

PR #58, opened 2018-04-04
Viite 1054 automatic removal of road name

plourenco90:
Needs to be merged after the VIITE-1029


skosonen:
"Do not merge" label can be removed after VIITE-1029 is merged in master.

PR #64, opened 2018-04-11
Viite 1247 the user can not open project

pedrosag:
Fixed the issue of the project not opening.
The underline cause was the un-use of VVH Geometry in the reverting of links, it now uses that.

Fixed tests for it.

Tests are passing correctly, check jenkins build 723.


pedrosag:
After merging with master Jenkins tests were passing as well, check build 726.

PR #67, opened 2018-04-13
Viite 1246 reversing direction with user given end addr m value breaks road lengths

Campanha:
Fixed links/calibrations points MValues and addrMValues recalculation for when user adds one/multiple new endAddrMValues for any link of the road part, despite the track it belongs to.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/CalibrationPointDAO.scala
```diff
@@ -51,8 +51,20 @@ object CalibrationPointDAO {
...
+    Q.queryNA[(Long, Long, Long, Double, Long)](baseQuery).list.map {
+      case (id, linkId, prId, linkM, addressM) => UserDefinedCalibrationPoint(id, linkId, prId, linkM, addressM)
+    }
+  }
+
+  def findEndCalibrationPoint(projectLinkId: Long, projectId: Long): Seq[UserDefinedCalibrationPoint] = {
+    val baseQuery =
+      s"""
+         Select ID, PROJECT_LINK_ID, PROJECT_ID, LINK_M, ADDRESS_M From CALIBRATION_POINT Where
+         PROJECT_LINK_ID = $projectLinkId And PROJECT_ID = $projectId And ADDRESS_M > 0
```
pedrosag: Offering a sugestion, if you want to get the calibration point at the END of a road wouldn't it make MORE sense to do something like:

Select ID, PROJECT_LINK_ID, PROJECT_ID, LINK_M, ADDRESS_M From CALIBRATION_POINT Where PROJECT_LINK_ID = $projectLinkId And PROJECT_ID = $projectId And ADDRESS_M IN (_Select Max(Address_M) rom CALIBRATION_POINT Where PROJECT_LINK_ID = $projectLinkId And PROJECT_ID = $projectId)_

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -133,22 +138,32 @@ object ProjectSectionCalculator {
     def makeStartCP(projectLink: ProjectLink) = {
       Some(CalibrationPoint(projectLink.linkId, if (projectLink.sideCode == TowardsDigitizing) 0.0 else projectLink.geometryLength, projectLink.startAddrMValue))
     }
+
     def makeEndCP(projectLink: ProjectLink, userDefinedCalibrationPoint: Option[UserDefinedCalibrationPoint]) = {
```
pedrosag: Methods makeEndCP and makeEndCPAtStartBorder are equal code wise, should they differ?

Campanha: they should not. . It was duplicated, glad you noticed that

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/CalibrationPointDAO.scala
```diff
@@ -51,8 +51,20 @@ object CalibrationPointDAO {
...
        """
 
-    Q.queryNA[(Long, Long, Long, Double, Long)](baseQuery).list.map{
-      case (id,linkId, prId, linkM, addressM) => UserDefinedCalibrationPoint(id,linkId, prId, linkM, addressM)
+    Q.queryNA[(Long, Long, Long, Double, Long)](baseQuery).list.map {
+      case (id, linkId, prId, linkM, addressM) => UserDefinedCalibrationPoint(id, linkId, prId, linkM, addressM)
+    }
+  }
+
+  def findEndCalibrationPoint(projectLinkId: Long, projectId: Long): Seq[UserDefinedCalibrationPoint] = {
```
skosonen: Will this work also with reversed project_links? I tried to test this sql in devtest but the only project having reversed links have only one calibration_point per project_link. This might be correct but I don't understand this yet well enough.

SELECT * FROM CALIBRATION_POINT WHERE project_id = 9070648;

Campanha: i will check that project case

skosonen: Might be good to create project where there is a reversed link with multiple calibration points?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/CalibrationPointDAO.scala
```diff
@@ -117,6 +129,12 @@ object CalibrationPointDAO {
       """.execute
   }
 
+  def removeAllCalibrationPoints(projectLinkIds: Set[Long]) = {
```
skosonen: Might be good to handle the special case when the projectLinkIds is an empty Set.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/CalibrationPointDAO.scala
Campanha: The Calibration_Point table defines if and where there is one new calibration point defined by the user for one project_link. 
"New endAddressMValue defined by user" logic indicates that the user can only define one new endAddress and not startAddress for one projec_link. (UI should not let user change the startAddress for one link even tough the ALUSSA field in form is not being blocked yet).
However for the query and the recalculation only the endAddress calibration point defined by user matters, thats why we only pick the last one from that project_link.

skosonen: Ok, so this should not be a problem.

skosonen: From this "...Select Max(Address_M)..." I thought that there could be more than one calibration points.

PR #73, opened 2018-04-24
VIITE-1287

pedrosag:
Removed the startDate validation and created a test for it.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -226,8 +226,7 @@ object RoadAddressDAO {
           terminated = '0' and
           valid_from <= sysdate and
           (valid_to is null or valid_to > sysdate) and
-          (end_date is null or end_date > sysdate) and
-          (start_date is null or start_date <= sysdate)
+          (end_date is null or end_date > sysdate)
```
hestrompa: Need to be careful, because this is not the only query that contains start_date and I think that we should test another cases where we are using start_date of road_addresses

pedrosag: Checked them but didn't find anything that should compromise other functions OR needed to be removed (well except a function that was not used anywhere).

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1431,23 +1438,6 @@ object RoadAddressDAO {
     query + s" WHERE ra.road_number = $road $trackFilter AND ra.floating = 0 " + withValidatyCheck
   }
 
-  def withRoadAddressSinglePart(roadNumber: Long, startRoadPartNumber: Long, track: Int, startM: Long, endM: Option[Long], optFloating: Option[Int] = None)(query: String): String = {
```
pedrosag: Removed for it was not being used anywhere.

pedrosag:
All tests are passing on Jenkins, check build 777 for them.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
skosonen: This could return multiple rows if we have planned road_addresses in to the future. "or end_date > sysdate" must be removed. We want to see only the most recent version even though it would be in the future.

PR #74, opened 2018-04-24
Viite 1289 when user redirect new road, viite changes continuity code of other road

pedrosag:


pedrosag:
All tests are passing, check Jenkins build 780.

PR #80, opened 2018-05-02
VIITE-1309 Logs accessible to monitoring

skosonen:
Deploy script makes symbolic link from “./logs” to “/data1/viite/logs”.


skosonen:
I might change the path to "/data1/logs/viite"

skosonen:
I have not yet tested this script so not sure if it works...

PR #92, opened 2018-05-11
VIITE-1325 Roads can't be selected after cancel button in an empty project

tomixii:
Before selection of a new road and canceling immediately in a fresh new project, user wasn't able to select a new road without restarting the project. Also the cancel button worked unlogically. Now these bugs are fixed and hopefully my fixes didn't create new bugs.

### DIFF
--- a/viite-UI/src/view/ProjectEditForm.js
+++ b/viite-UI/src/view/ProjectEditForm.js
@@ -371,21 +371,14 @@
       };
 
       var cancelChanges = function() {
-        selectedProjectLinkProperty.setDirty(false);
-        if(projectCollection.isDirty()) {
-          projectCollection.revertLinkStatus();
-          projectCollection.setDirty([]);
-          projectCollection.setTmpDirty([]);
-          projectLinkLayer.clearHighlights();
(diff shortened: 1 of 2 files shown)


viite-UI/src/view/ProjectEditForm.js
```diff
@@ -371,21 +371,14 @@
...
 
       var cancelChanges = function() {
-        selectedProjectLinkProperty.setDirty(false);
-        if(projectCollection.isDirty()) {
-          projectCollection.revertLinkStatus();
-          projectCollection.setDirty([]);
-          projectCollection.setTmpDirty([]);
-          projectLinkLayer.clearHighlights();
-          $('.wrapper').remove();
-          formCommon.toggleAdditionalControls();
```
skosonen: So this is not needed any more? Or did this drop out by mistake?

tomixii: I dropped it because it does toggleAdditionalControls() in the reOpenCurrent function which is called on line 381. Viite still works fine, now it just calls toggleAdditionalControls() one time less.

PR #94, opened 2018-05-11
VIITE-1330 Unnecessary popup window after cancel in edit project view

tomixii:
There was an unnecessary popup window after pressing cancel button in edit project view when user hasn't made any edits. Now the popup shows up only when there is something to save and user presses cancel button.

### DIFF
--- a/viite-UI/src/view/ProjectEditForm.js
+++ b/viite-UI/src/view/ProjectEditForm.js
@@ -371,21 +371,14 @@
       };
 
       var cancelChanges = function() {
-        selectedProjectLinkProperty.setDirty(false);
-        if(projectCollection.isDirty()) {
-          projectCollection.revertLinkStatus();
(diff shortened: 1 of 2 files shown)


viite-UI/src/view/ProjectEditForm.js
```diff
@@ -371,21 +371,14 @@
...
 
       var cancelChanges = function() {
-        selectedProjectLinkProperty.setDirty(false);
-        if(projectCollection.isDirty()) {
-          projectCollection.revertLinkStatus();
-          projectCollection.setDirty([]);
-          projectCollection.setTmpDirty([]);
-          projectLinkLayer.clearHighlights();
-          $('.wrapper').remove();
-          formCommon.toggleAdditionalControls();
```
skosonen: Is this toggleAdditionalControls left out on purpose?

PR #95, opened 2018-05-11
Viite 1278 1265 track code changes and fixes

leirasf:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -230,6 +282,7 @@ object ProjectSectionCalculator {
               assignCalibrationPoints(Seq(), sec.left.links, calMap)
           }
         }
+        //validateCalibrationPointsDiscontinuityOnTracks()
```
skosonen: Should this be removed or taken in use?

pedrosag: I second that question.

leirasf: Well this method is not used at the moment because we didn't decide with Janne how it should behave. We should remove the usage and then add some comment to method, so we can reuse it when we have a more clear idea of want we need to do there.

leirasf: I will handle it.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -46,28 +47,36 @@ object CommonHistoryFiller {
     }.toSeq ++ newRoadAddresses.filterNot(_.id == NewRoadAddress)
   }
 
-  private def applyTransfer(currentRoadAddresses : Seq[RoadAddress])(projectLinks: Seq[ProjectLink], newRoadAddresses: Seq[RoadAddress]) :Seq[RoadAddress]={
+  private def applyTransfer(currentRoadAddresses: Seq[RoadAddress])(projectLinks: Seq[ProjectLink], newRoadAddresses: Seq[RoadAddress]): Seq[RoadAddress] = {
     val transferredLinks = projectLinks.filter(_.status == LinkStatus.Transfer)
+    val (trackChangedLinks, trackUnchangedLinks) = transferredLinks.partition(pl => {
+      val currentRoadAddress = currentRoadAddresses.find(ra => ra.id == pl.roadAddressId)
+      currentRoadAddress.nonEmpty && ((currentRoadAddress.get.track == Combined && pl.track != Combined) || (currentRoadAddress.get.track != Combined && pl.track == Combined))
```
Campanha: we could use currentRoadAddress.get.track != pl.track instead

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
pedrosag: All right, thank you :)

PR #96, opened 2018-05-11
Viite 1242 road types stay same in road address change v2

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -109,6 +121,56 @@ object CommonHistoryFiller {
...
+      }
+    }
+
+    val startNeedsCP = roadAddresses.head
+    val endNeedsCP = roadAddresses.last
+
+    val returnObject = roadAddresses.length match {
+      case 2 => {
+        Seq(fillCPs(startNeedsCP, putCPAtStart = true)) ++ Seq(fillCPs(endNeedsCP))
+      }
```
leirasf: This case is not needed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -109,6 +121,56 @@ object CommonHistoryFiller {
...
+      } else {
+        if (!putCPAtStart) {
+          roadAddress.copy(calibrationPoints = (None, endCP))
+        } else {
+          roadAddress.copy(calibrationPoints = (startCP, None))
+        }
+      }
+    }
+
+    val startNeedsCP = roadAddresses.head
```
leirasf: Did you tests the case where you have a empty roadAddresses. I think you will get the following error "head of empty list".

pedrosag:
@leirasf I addressed all the issues you raised, can you re-checked them?

PR #99, opened 2018-05-14
Viite 1263 apply changes fix

Campanha:
Changes should changes since last change timedate.


Campanha:
need yet to remove logs and others

PR #102, opened 2018-05-15
VIITE-1332 creation of projects with same name

tomixii:
User can't create or save a project with same name as another project in the database. It is NOT case sensitive so for example "ProjectOne" equals "projectone". If the user is trying to save project with same name, popup-window appears and says: "Nimellä (project name) on jo olemassa projekti. Muuta nimeä."("There's already a project with this name. Change the name, please.")


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -933,6 +933,17 @@ object ProjectDAO {
     dateTime.map(dt => new Timestamp(dt.getMillis))
   }
 
+  def uniqueName(projectId: Long, projectName: String): Boolean = {
+    val query =
+      s"""
+         SELECT *
+         FROM project
+         WHERE UPPER(name)=UPPER('$projectName') and state<>7
```
oappicgi: case (upper) shouldn't matter in SQL
To make query bit faster add "ROWNUM = 1" clause so if it finds one it stops searching

tomixii: UPPER actually seems to matter here. I'll add that "ROWNUM=1".

PR #106, opened 2018-05-16
VIITE-1334

pedrosag:
Added additional logging messages on multiple steps.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1581,6 +1607,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       logger.error(s" There are no roadlinks to update  with names, rollbacking update ${project.id}")
```
Campanha: this log should also change as should say something like "..there is no road addresses to update, rollbacking update.."

PR #108, opened 2018-05-16
VIITE-1242

pedrosag:
Added new historic rows on the road address table prior to expiring.


skosonen:
ProjectServiceSpec "process roadChange data and expire the roadLink" -test fails.

pedrosag:
Corrected the faulty test by adding behaviour to deal with the newly created historic links.
One can check jenkins build 911 for confirmation.

PR #109, opened 2018-05-18
Viite 1336 roadnames table gets valid duplicate roadnames for same roadnumber

Campanha:


digiroad2-oracle/src/main/resources/db/migration/V0_8__create unique_constraint_project_link_name.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "PROJECT_LINK_NAME" ADD CONSTRAINT "PROJECT_LINK_NAME_ROW_UNIQUE" UNIQUE ("PROJECT_ID", "ROAD_NUMBER")
+USING INDEX PCTFREE 10 INITRANS 2 MAXTRANS 255 COMPUTE STATISTICS
```
skosonen: I think these should be removed and keep only the first row:
`ALTER TABLE "PROJECT_LINK_NAME" ADD CONSTRAINT "PROJECT_LINK_NAME_ROW_UNIQUE" UNIQUE ("PROJECT_ID", "ROAD_NUMBER");`

PR #110, opened 2018-05-18
VIITE-1315 Project name and start date are missing

tomixii:
After reserving road part and deleting the reservation project's name and date disappeared. This bug is now fixed.


viite-UI/src/view/ProjectForm.js
```diff
@@ -650,13 +653,33 @@
...
+        if (currentProject.isDirty) {
+            new GenericConfirmPopup('Haluatko tallentaa tekemäsi muutokset?', {
+                successCallback: function () {
+                    if (!disabledInput) {
+                        createOrSaveProject();
+                        eventbus.once('roadAddress:projectSaved', function () {
+                            _.defer(function () {
+                                closeProjectMode(true);
+                            });
+                        });
```
leirasf: If it's the `createOrSaveProject `that triggers the `roadAddress:projectSaved` it should be registed before we call method.

PR #114, opened 2018-05-21
Viite 1339 roadaddress direction correction

oappicgi:
we basically  took direction from vvh trafic direction which could be wrong. Now we use unknown unless we have direction in our database.

Added test for having road in db and changed existing test end result to have expected outcome


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressLinkBuilder.scala
```diff
@@ -110,14 +110,9 @@ object RoadAddressLinkBuilder extends AddressLinkBuilder {
...
-      SideCode.TowardsDigitizing
-    } else if (roadLink.trafficDirection == TrafficDirection.AgainstDigitizing) {
-      SideCode.AgainstDigitizing
-    } else if (roadLink.trafficDirection == TrafficDirection.BothDirections) {
-      SideCode.BothDirections
-    } else {
-      SideCode.Unknown
+    val sideCode =headAddress match {
+      case Some(road) => road.sideCode
+      case _ => SideCode.Unknown
```
Campanha: in case there is no address(es), shouldnt we then get the sideCode from roadLinks?

oappicgi: vvhroadlink doesnt have sidecode. It has trafic direction but according to JanneG it is very unreliable and it is better to have unknown than bad data

PR #118, opened 2018-05-22
VIITE-1350 fixing suravage links to UI

hestrompa:


oappicgi:
one UI test is failing

PR #124, opened 2018-05-24
VIITE-1341 current calibration points should be preserved in project

tomixii:
Viite doesn't lose current calibration points when user does transfer operation. Please tell me if the code quality needs to be improved or you notice I have used wrong functions to implement the functionality.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1529,4 +1529,11 @@ object RoadAddressDAO {
           valid_from <= sysdate and floating = 0 and ely = $ely"""
     queryList(query)
   }
+
+  def getLinkCPValue(roadAddressId: Long): Long = {
+    val query = s"""SELECT ra.calibration_points
+                    FROM road_address ra
+                    WHERE ra.id=$roadAddressId AND ROWNUM=1"""
```
skosonen: road_address table's id-field is already unique, so " AND ROWNUM=1" can be removed.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -160,29 +160,33 @@ object ProjectSectionCalculator {
 
     def assignCalibrationPoints(ready: Seq[ProjectLink], unprocessed: Seq[ProjectLink],
                                 calibrationPoints: Map[Long, UserDefinedCalibrationPoint]): Seq[ProjectLink] = {
+      val link = unprocessed.head
+      //checks calibration points which the link had before starting the project
+      val raCPValue = RoadAddressDAO.getLinkCPValue(link.roadAddressId)
+      val raStartCP = raCPValue == CalibrationCode.AtBeginning.value || raCPValue == CalibrationCode.AtBoth.value
```
skosonen: Here you could compare CalibrationCodes instead of Long values.

skosonen:
Unit tests should be updated. At least this one has many failing tests now: ProjectSectionCalculatorSpec.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
oappicgi: that works. You could have just used queryById and retrive the whole roadaddress data, but that should be slightly faster since it doesn't have to map so many values. the thing is that if someone later wants to change this code and asumes 0 value means no calibration point as it should, but in orelse it could actually mean also that it could not find previous roadaddress.

tomixii: What value would be the best to return or what should happen if a road address could not be found from the database?

pedrosag:
Tests pass, check build 949 on jenkins.

PR #127, opened 2018-05-25
Viite 1340 check what happen in road address change when road links has been changed

pedrosag:


skosonen:
We should treat rows with valid_to -value as deleted. We should never try to find in code rows that have valid_to value set.
If it seems that we need any rows with valid_to value set, something went wrong already earlier.

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -383,7 +383,7 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
   get("/roadlinks/roadaddress/project/all/projectId/:id") {
     val projectId = params("id").toLong
     try {
-      projectService.getRoadAddressSingleProject(projectId) match {
+      projectService.getRoadAddressSingleProject(projectId, false) match {
```
leirasf: Use like this `projectService.getRoadAddressSingleProject(projectId, withActiveRoadAddress = false)` to be easier to read

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -522,13 +522,17 @@ object ProjectDAO {
...
+      if (withActiveRoadAddresses) {
+        ("AND RA.END_DATE IS NULL AND RA.VALID_TO IS NULL", "AND RA.END_DATE IS NULL AND RA.VALID_TO IS NULL")
+      } else ("", "")
     val sql =
       s"""
         SELECT id, road_number, road_part_number, length, length_new,
           ely, ely_new,
           (SELECT DISCONTINUITY FROM ROAD_ADDRESS ra WHERE ra.road_number = gr.road_number AND
-            ra.road_part_number = gr.road_part_number AND RA.END_DATE IS NULL AND RA.VALID_TO IS NULL
+            ra.road_part_number = gr.road_part_number $activeDiscontinuity
```
leirasf: This will return more than one row, when you start having history.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -551,8 +555,7 @@ object ProjectDAO {
               LEFT JOIN
               ROAD_ADDRESS ra ON ((ra.road_number = rp.road_number AND ra.road_part_number = rp.road_part_number) OR ra.id = pl.ROAD_ADDRESS_ID)
               WHERE
-                rp.project_id = $projectId AND
-                RA.END_DATE IS NULL AND RA.VALID_TO IS NULL
+                rp.project_id = $projectId $activeMain
```
leirasf: I don't think you should have change this query.

pedrosag:
Tests are passing check build 967 on Jenkins.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1214,12 +1214,18 @@ object RoadAddressDAO {
...
-      else
-        ""
-      val query =
-        s"""
+    val historyFilter = if (includeHistory)
+      "AND end_date is null"
+    else
+      ""
+
+    val validToFilter = if (rejectInvalids)
```
skosonen: I think this "valid_to is null" check should be always in query. We should treat the rows as deleted if valid_to is set.

skosonen:
@pedrosag Okay! Sorry, now I also understand the reason for that. 😊

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1558,12 +1558,13 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       ).getOrElse(roadAddress.startAddrMValue, roadAddress.endAddrMValue, roadAddress.startMValue, roadAddress.endMValue)
     }
 
-    split.flatMap(pl =>
+    split.flatMap(pl => {
+      val floatingValue = roadAddress.validTo.isDefined && roadAddress.validTo.get.isBeforeNow
       pl.status match {
```
skosonen: I think it should be enough to check that valid_to is just set. We should never set valid_to date to the future so this isBeforeNow is not needed.

PR #128, opened 2018-05-28
VIITE-1238 Move the change table only from title bar

tomixii:
Added a simple line which let's the user to move the change table only from the header.


oappicgi:
nice find, thought that was bigger task

PR #129, opened 2018-05-28
VIITE-1315 Clears text inputs after successful road reservation

tomixii:
Road reservation clears text input fields( tie, aosa, losa) if the reservation was successful(road is not reserved for any other project or road address doesn't exist).


tomixii:
Yes there's outdated information in the VIITE-1315 description. We discussed about this with Janne and Henni and we decided that it should clear the fields already after successful reserve.

PR #130, opened 2018-05-28
Viite 1354 1356 1357 wrong project link calculator

leirasf:
The start address measure validations were removed from unit tests because the average rule was dopped for project delta calculator.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -113,6 +113,13 @@ case class ProjectLink(id: Long, roadNumber: Long, roadPartNumber: Long, track:
       case _ => throw new InvalidAddressDataException(s"Bad sidecode $sideCode on project link")
     }
   }
+
+  def addrMLenght() = {
```
hestrompa: Should be "length" instead of "lenght"

PR #131, opened 2018-05-28
VIITE-1361

pedrosag:
Made a simple method to return the name of the road that is being split.

The origin is firstly the project link, then the road address, if none if found from these 2 sources then we return the original, un-modified value of the name that resides in the project link.

Also corrected a little bug, resulting of calling a .map() method on a empty collection.

Added the road name parameter to the list of outgoing values.

### DIFF
--- a/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
+++ b/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
@@ -980,7 +980,8 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
             "status" -> splittedLinks.status.value,
             "roadTypeId" -> splittedLinks.roadType.value,
             "discontinuity" -> splittedLinks.discontinuity.value,
-            "elyCode" -> splittedLinks.ely
+            "elyCode" -> splittedLinks.ely,
+            "roadName" -> splittedLinks.roadName.getOrElse("")
           ))
       }
       case LinkStatus.Terminated => {
@@ -994,7 +995,8 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
             "status" -> splittedLinks.status.value,
             "roadTypeId" -> splittedLinks.roadType.value,
             "discontinuity" -> splittedLinks.discontinuity.value,
-            "elyCode" -> splittedLinks.ely
+            "elyCode" -> splittedLinks.ely,
+            "roadName" -> splittedLinks.roadName.getOrElse("")
           ))
       }
       case _ => {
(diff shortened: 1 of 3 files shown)


Campanha:
you should also refactor the splitToApi mapping since the map values are same and only key differs

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -531,6 +531,27 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+      * Fetches the projectLink name, first from the project link, if that's not available then search for the road address.
+      *
+      * @param projectLink
+      * @return
+      */
+    def fillRoadNames(projectLink: ProjectLink): ProjectLink = {
+      val projectLinkName = ProjectLinkNameDAO.get(projectLink.roadNumber, projectLink.projectId)
+      val trueRoadName = if (projectLinkName.isEmpty) {
+        val roadName = RoadNameDAO.getLatestRoadName(projectLink.roadNumber)
+        if (roadName.isEmpty) {
```
hestrompa: I'm sure we have a better way to validate this and reduce the amount of code just to make sure that roadname is not empty. You can use headOption or  .map(_.roadName).getOrElse(""). All this "if else if else" code can be suspicious in the future

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -1008,7 +1010,8 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
             "status" -> splittedLinks.status.value,
             "roadTypeId" -> splittedLinks.roadType.value,
             "discontinuity" -> splittedLinks.discontinuity.value,
-            "elyCode" -> splittedLinks.ely
+            "elyCode" -> splittedLinks.ely,
+            "roadName" -> splittedLinks.roadName.getOrElse("")
```
hestrompa: The value here I believe it shouldn't be an Option[String] anymore. If we handle everything on service level should be better

hestrompa:
A test for this would be nice

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -540,7 +561,13 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
           splitOptions
       val r = preSplitSuravageLinkInTX(linkId, userName, updatedSplitOptions)
```
leirasf: Please, instead of `val r = blabla` use `val (val1, val2, val3) = blablabla` to be easier to read.

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
pedrosag: Would require changing the whole ProjectLink object in order to do it, making it a permanent String object then it would be detrimental in the cases that the project links has no name.

pedrosag:
All tests are passing, check build 970 on Jenkins.

PR #132, opened 2018-05-29
VIITE-1364 Too long project names break Viite

tomixii:
Added a function which cuts the name if it is over 20 characters long and adds "..." in the end. This is because earlier long names broke "Poistu projektista" and "Poista projekti" buttons.

### DIFF
--- a/viite-UI/src/view/FormCommon.js
+++ b/viite-UI/src/view/FormCommon.js
@@ -10,16 +10,21 @@
     };
 
     var titleWithProjectName = function(projectName, project) {
-      return '<span class ="edit-mode-title">'+projectName+'<button id="editProject_'+ project.id +'" ' +
+      return '<span class ="edit-mode-title">' + cutProjectName(projectName) + '<button id="editProject_' + project.id + '" ' +
         'class="btn-edit-project" style="visibility:hidden;" value="' + project.id + '"></button></span>' +
         '<span id="closeProjectSpan" class="rightSideSpan" style="visibility:hidden;">Poistu projektista</span>';
     };
 
-      var addRoadNameField = function (name, isBlocked) {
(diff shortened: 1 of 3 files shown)


viite-UI/src/view/FormCommon.js
```diff
@@ -10,16 +10,21 @@
...
         '<span id="closeProjectSpan" class="rightSideSpan" style="visibility:hidden;">Poistu projektista</span>';
     };
 
-      var addRoadNameField = function (name, isBlocked) {
-          var nameToDisplay = _.isUndefined(name) || _.isNull(name) || name === 'null' || name === '' ? "" : name;
-          var disabled = nameToDisplay !== "" && isBlocked;
-          return '<input type="text" class="form-control" style="float:none; display:inline-block" id = "roadName" value="' + nameToDisplay + '" ' + (disabled ? 'disabled' : '') + '/>';
-      };
+    var cutProjectName = function (name) {
+      var maxNameLength = 20;
```
hestrompa: In the task description it says that the limit length is 32. Should we cut it at 20 characters?

tomixii: Yes, the maximum length for project name is 32 characters at the moment. It would be maybe better to make it less but in the other hand 20 characters seems a bit too little. One improvement could be that instead of cutting the name based on letter count we could cut the name based on it's pixel width on screen since only the names written in caps lock are the problem here.

Campanha: If that is the case, the VARCHAR2(32 BYTE) in "NAME" collumn for PROJECT table doesnt guarantee that length is 32 chars long.
Google says that in UTF-8 characters with a code less than 128 will require 1 byte only, and the next 1920 character codes require 2 bytes only. (for e.g. chars with diaeresis require 2 bytes)

PR #134, opened 2018-05-29
VIITE-1257 Starttin date as a column header in project list table

tomixii:
Fixed label widths so I could fit one more column.


viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -103,6 +105,11 @@
         });
       };
 
+      var cutProjectName = function (name) {
+        var maxNameLength = 24;
```
hestrompa: Same in VIITE-1364 but different lengths? :/

tomixii: There was more space for the name here but I can change them both to be same.

PR #139, opened 2018-05-31
VIITE-1368 Road name is not obligatory if road number is greater than 70000

tomixii:
If road number is greater than 70000 that means the road is for pedestrian and thus doesn't need a name.

### DIFF
--- a/viite-UI/src/view/ProjectEditForm.js
+++ b/viite-UI/src/view/ProjectEditForm.js
@@ -121,8 +121,9 @@
       var rootElement = $('#feature-attributes');
       var inputs = rootElement.find('input');
       var filled = true;
+      var pedestrianRoads = 70000;
       for (var i = 0; i < inputs.length; i++) {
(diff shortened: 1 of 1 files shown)


viite-UI/src/view/ProjectEditForm.js
```diff
@@ -122,7 +122,8 @@
       var inputs = rootElement.find('input');
       var filled = true;
       for (var i = 0; i < inputs.length; i++) {
-        if (inputs[i].type === 'text' && !inputs[i].value) {
+        var isPedestrian = $('#tie')[0].value >= 70000; // roads with road number over 70000 are pedestrian roads
```
oappicgi: we rather not have magic number, even though comment is nice, but it is better to initialize pedestrianroads variable/value and use that. Then incase someone decides pedestrian roads start at 70000 it is easier to fix

tomixii:
Fixed!

PR #141, opened 2018-06-01
Viite 1344 reverting previous changes

hestrompa:


viite-UI/src/view/LinkPropertyForm.js
```diff
@@ -764,7 +776,9 @@
 
       eventbus.on('adjacents:roadTransfer', function(result, sourceIds, targets) {
         $('#additionalSource').remove();
-        $('#adjacentsData').remove();
+        $('#control-label-floating').remove();
+        $('#adjacentsData').empty();
+        //$('#adjacentsData').remove();
```
leirasf: Remove comment code.

viite-UI/src/view/LinkPropertyForm.js
```diff
@@ -793,27 +807,34 @@
...
           floatingPart = floatingPart + additionalSource(fr.linkId, fr.marker);
         });
-        $(".form-group:last").after(floatingPart);
-        $('[id*="additionalSourceButton"]').click(floatingRoads,function(event) {
-          processAdditionalFloatings(floatingRoads,event.currentTarget.value);
-        });
+        if(floatingRoads.length === 0){
+          applicationModel.setContinueButton(true);
+          rootElement.find('.link-properties button.continue').attr('disabled', false);
+          //eventbus.trigger('adjacents:startedFloatingTransfer');
```
leirasf: Remove comment code

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/AddressLinkBuilder.scala
```diff
@@ -213,7 +214,7 @@ trait AddressLinkBuilder {
           right.map(_.copy(segmentMValue = if (nextSegment.sideCode == SideCode.AgainstDigitizing) startMValue else endMValue)))
       }
 
-      if(nextSegment.sideCode.value != previousSegment.sideCode.value)
+      if(nextSegment.sideCode.value != previousSegment.sideCode.value && GeometryUtils.geometryLength(previousSegment.geometry) != 0 && GeometryUtils.geometryLength(nextSegment.geometry) != 0)
```
Campanha: should we change the message to log something about geometry length being zero?

hestrompa: Don't think so. This is just to prevent the cases that we have only one point in geometry, and then it gets random sidecode value, since it's impossible to build a direction of vector with just one point

hestrompa:
I'm just fixing one little issue before merge

PR #144, opened 2018-06-04
Viite 1369 change apply changes batch logic

Campanha:
Changes to current and history should be independent

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapper.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapper.scala
@@ -153,10 +153,10 @@ object RoadAddressChangeInfoMapper extends RoadAddressMapper {
     preTransferCheckBySection(originalAddressSections)
     val groupedChanges = changes.groupBy(_.vvhTimeStamp).values.toSeq
     val appliedChanges = applyChanges(groupedChanges.sortBy(_.head.vvhTimeStamp), roadAddresses.mapValues(_.allSegments))
-    val currentSectionChanges = groupByCurrentRoadSection(currentSections, appliedChanges.values.map(
-      s => LinkRoadAddressHistory(s.partition(_.endDate.isEmpty))))
-    val historySectionChanges = groupByHistoryRoadSection(historySections, appliedChanges.values.map(
-      s => LinkRoadAddressHistory(s.partition(_.endDate.isEmpty))))
+    val mappedChanges = appliedChanges.values.map(
+      s => LinkRoadAddressHistory(s.partition(_.endDate.isEmpty)))
+    val currentSectionChanges = groupByCurrentRoadSection(currentSections, mappedChanges)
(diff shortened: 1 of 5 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapper.scala
```diff
@@ -145,19 +145,34 @@ object RoadAddressChangeInfoMapper extends RoadAddressMapper {
...
     preTransferCheckBySection(originalAddressSections)
     val groupedChanges = changes.groupBy(_.vvhTimeStamp).values.toSeq
     val appliedChanges = applyChanges(groupedChanges.sortBy(_.head.vvhTimeStamp), roadAddresses.mapValues(_.allSegments))
-    val result = postTransferCheckBySection(groupByRoadSection(sections, appliedChanges.values.map(
-      s => LinkRoadAddressHistory(s.partition(_.endDate.isEmpty)))), originalAddressSections)
+    val mappedChanges = appliedChanges.values.map(
+      s => LinkRoadAddressHistory(s.partition(_.endDate.isEmpty)))
+    val currentSectionChanges = groupByCurrentRoadSection(currentSections, mappedChanges)
+    val historySectionChanges = groupByHistoryRoadSection(historySections, mappedChanges)
+    val result = postTransferCheckBySection(currentSectionChanges, historySectionChanges, originalAddressSections, historyAddressSections)
```
leirasf: Instead of having all of that duplicated code, shouldn't we have a separated method that then can be called to current sections and then by history sections?

Campanha: They are actually different methods. But i will still create one method to call them both.

Campanha: I will also create some tests for it

Campanha:
all tests+new passing

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressMapper.scala
```diff
@@ -81,38 +81,58 @@ trait RoadAddressMapper {
...
+    postTransferChecks((s._1, s._2.flatMap(_.currentSegments), Seq()))
   }
 
-  def postTransferChecksWithHistory(s: (RoadAddressSection, Seq[LinkRoadAddressHistory])): Unit = {
-    postTransferChecks((s._1, s._2.flatMap(_.currentSegments)))
+  def postTransferChecksForHistory(s: (RoadAddressSection, Seq[LinkRoadAddressHistory])): Unit = {
+    postTransferChecks((s._1, Seq(), s._2.flatMap(_.historySegments)))
   }
 
   protected def commonPostTransferChecks(seq: Seq[RoadAddress], addrMin: Long, addrMax: Long): Unit = {
```
oappicgi: could we rename seq to something bit more descriptive.

Campanha:
LogUtils.time import not working/warnings

PR #152, opened 2018-06-07
Viite 1377 splitted link does not find geometry

pedrosag:
Tests are passing, check Jenkins build 977.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
@@ -4,6 +4,7 @@ import java.net.ConnectException
 
 import fi.liikennevirasto.digiroad2._
 import fi.liikennevirasto.digiroad2.asset._
+import fi.liikennevirasto.digiroad2.client.vvh.FeatureClass.AllOthers
 import fi.liikennevirasto.digiroad2.client.vvh.{ChangeInfo, ChangeType, VVHHistoryRoadLink, VVHRoadlink}
 import fi.liikennevirasto.digiroad2.linearasset.{RoadLink, RoadLinkLike}
 import fi.liikennevirasto.digiroad2.oracle.OracleDatabase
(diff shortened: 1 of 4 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -190,7 +215,9 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
     val returningTopology = filledTopology.filter(link => !complementaryLinkIds.contains(link.linkId) ||
       complementaryLinkFilter(link))
 
-    returningTopology ++ missingFloating.map(floating => floating.copy(roadLinkType = RoadLinkType.FloatingRoadLinkType))
+
+    val test = missingFloating.map(floating => floating.copy(roadLinkType = RoadLinkType.FloatingRoadLinkType))
```
oappicgi: bit more descriptive name to test

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -81,13 +82,13 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
         roadLinkService.getRoadLinksHistoryFromVVH(floatingAddresses.map(_.linkId).toSet)
       }
     }
-    val historyLinkAddresses = time(logger, "Build history link addresses") {
+    val historyFloatingLinkAddresses = time(logger, "Build history link addresses") {
```
hestrompa: I think that we should not change this kind of names from "history link addresses" to "history floating addresses", because we can have history links also without being floating.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
Campanha: this lines can be reverted to previous version

pedrosag:
Tests are passing on Jenkins, check build 985.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/package.scala
```diff
@@ -151,6 +154,13 @@ package object viite {
     }.toSeq
   }
 
+  def featureClassCodeToFeatureClass: Map[Int, FeatureClass] = Map(
+    12316 -> FeatureClass.TractorRoad,
+    12141 -> FeatureClass.DrivePath,
+    12314 -> FeatureClass.CycleOrPedestrianPath,
+    12312 -> FeatureClass.WinterRoads
+  )
+
```
leirasf: This is duplicated it shouldn't be used outside vvhClient

PR #154, opened 2018-06-07
VIITE-1385 Unable to save unchanged roadlinks

tomixii:
User was unable to save unchanged road link at road number 73 and road part number 18. Now it is fixed but some regression testing is needed.

### DIFF
--- a/viite-UI/src/model/ProjectCollection.js
+++ b/viite-UI/src/model/ProjectCollection.js
@@ -329,7 +329,7 @@
         roadEly: Number(roadAddressProjectForm.find('#ely')[0].value),
         roadLinkSource: Number(_.first(changedLinks).roadLinkSource),
         roadType: Number(roadAddressProjectForm.find('#roadTypeDropDown')[0].value),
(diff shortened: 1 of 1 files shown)


Campanha:
Ive tested the new user given addr m value for new link after the unchanged saving fix and its not working now. I tested without 1385 changes locally and when ok.

Test case: selected on grey link. Action Uusi and save. Selected again and gave new end addr-m value. bigger than 5% of original road length. Warning pop up appeared.

viite-UI/src/model/ProjectCollection.js
```diff
@@ -316,8 +316,9 @@
 
       var projectId = projectInfo.id;
       var coordinates = applicationModel.getUserGeoLocation();
-      var roadAddressProjectForm = $('#roadAddressProjectForm');
-      var dataJson = {
+        var roadAddressProjectForm = $('#roadAddressProjectForm');
+        var endDistance = roadAddressProjectForm.find('#endDistance')[0];
```
skosonen: This could be simply:
`        var endDistance = $('#endDistance')[0];`

PR #155, opened 2018-06-07
Viite 1368 1316 fixing road names are not obligatory if road number is above 70000

pedrosag:
Tests are passing on Jenkins, check build 983.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -1702,7 +1702,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def handleNewRoadNames(projectLinks: Seq[ProjectLink], project: RoadAddressProject) = {
     val projectLinkNames = ProjectLinkNameDAO.get(projectLinks.map(_.roadNumber).toSet, project.id).filterNot(p => {
-      p.roadNumber > 70000 && p.roadName == null
+      p.roadNumber >= maxRoadNumberDemandingRoadName && p.roadName == null
     })
     val existingInRoadNames = projectLinkNames.flatMap(n => RoadNameDAO.getCurrentRoadNamesByRoadNumber(n.roadNumber)).map(_.roadNumber).toSet
     val (existingLinkNames, newLinkNames) = projectLinkNames.partition(pln => existingInRoadNames.contains(pln.roadNumber))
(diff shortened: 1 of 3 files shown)


pedrosag:
Build 984 has the tests, passing after the master merge.

pedrosag:
Thank you @tomixii was hungry and "ate" a 0 from the title :P

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/ProjectServiceSpec.scala
```diff
@@ -2278,4 +2278,23 @@ class ProjectServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
 
   }
 
+  test("road name should not be saved saved on TR success response if road number > 70.000") {
```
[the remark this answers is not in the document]
pedrosag: You are right I am gonna do that real quick like.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1674,7 +1674,9 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     logger.info(s"Found ${expiringRoadAddresses.size} to expire; expected ${replacements.map(_.roadAddressId).toSet.size}")
     ProjectDAO.moveProjectLinksToHistory(projectID)
     logger.info(s"Moving project links to project link history.")
-    handleNewRoadNames(projectLinks, project)
+    val validLinksToNames = projectLinks.filter(_.roadNumber <= maxRoadNumberDemandingRoadName)
+    if (validLinksToNames.nonEmpty)
+      handleNewRoadNames(validLinksToNames, project)
```
tomixii: Road number 70000 doesn't need name. Here road number 70001 is the first road which doesn't need name. Also the roads with number 70000 or over CAN be named but it's not mandatory and if they are named we should save the name to the database.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1703,7 +1701,9 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
   }
 
   def handleNewRoadNames(projectLinks: Seq[ProjectLink], project: RoadAddressProject) = {
-    val projectLinkNames = ProjectLinkNameDAO.get(projectLinks.map(_.roadNumber).toSet, project.id)
+    val projectLinkNames = ProjectLinkNameDAO.get(projectLinks.map(_.roadNumber).toSet, project.id).filterNot(p => {
+      p.roadNumber > 70000 && p.roadName == null
+    })
```
tomixii: > should be >= and I think we should use maxRoadNumberDemandingRoadName instead of 70000

pedrosag:
Implemented @Campanha suggestion as it makes way more sense, all tests are passing, check build 986 on Jenkins.

PR #156, opened 2018-06-08
VIITE-1262 Project list is now filtered also after sorting

tomixii:
Quick fix for filtering the project list also after sorting.


tomixii:
More fixes coming...

PR #157, opened 2018-06-08
VIITE-1259 Displaying old projects on the project list

tomixii:
Added checkbox which shows all projects sent to Tierekisteri. Project list doesn't show projects which have been sent to Tierekisteri over two days ago anymore.


viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -121,35 +122,52 @@
...
+      filterByUser();
     };
 
     function bindEvents() {
 
       eventbus.once('roadAddressProjects:fetched', function(projects) {
-        projectArray = projects;
-        createProjectList(projects);
+        projectArray = _.filter(projects, function(proj) {
+          return proj.statusCode !== 7; //filter deleted projects out
```
Campanha: you can add the Project status code to the LinkValues.js in order to avoid magic numbers code ^^

viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -121,35 +122,52 @@
...
+        });
+        createProjectList(projectArray);
       });
 
-      var createProjectList = function(projects) {
+      var createProjectList = function(projects, sortFunction = function(a,b) {return a.ely - b.ely}, order = 1) {
         var unfinishedProjects = _.filter(projects, function(proj) {
-          return (proj.statusCode >= 1 && proj.statusCode <= 5) || proj.statusCode === 8;
+          if (proj.statusCode === 5) {
```
Campanha: same as previous comment

viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -121,35 +122,52 @@
...
-        projectArray = projects;
-        createProjectList(projects);
+        projectArray = _.filter(projects, function(proj) {
+          return proj.statusCode !== projectStatus.Deleted.value; //filter deleted projects out
+        });
+        createProjectList(projectArray);
       });
 
-      var createProjectList = function(projects) {
+      var createProjectList = function(projects, sortFunction = function(a,b) {return a.ely - b.ely}, order = 1) {
```
Campanha: i got one error in compilation where it seems to warning that in current jshint the default parameters  (either functins or values) dont work.

viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -121,35 +122,52 @@
...
         var unfinishedProjects = _.filter(projects, function(proj) {
-          return (proj.statusCode >= 1 && proj.statusCode <= 5) || proj.statusCode === 8;
+          if (proj.statusCode === projectStatus.Saved2TR.value) {
+            var hoursInDay = 24;
+            var millisecondsToHours = 1000*60*60;
+            //check if show all TR projects checkbox is checked or the project has been sent to TR under two days ago
+            return $('#TRProjectsVisibleCheckbox')[0].checked || (new Date() - new Date(proj.dateModified.split('.').reverse().join('-'))) / millisecondsToHours < hoursInDay * 2;
+          }
+          return ((proj.statusCode >= 1 && proj.statusCode <= 4) || proj.statusCode === projectStatus.ErrorInViite.value);
```
Campanha: there are yet two missing project status codes ^^

tomixii:
Yes, I noticed those two missing project status codes but I thought that the code would have been more readable with status codes as integers rather than their description in that case. Now it says "true when project status is between Incomplete and TRProcessing" which I thought that didn't tell anything. Now, with the second thought I realized that between is pretty good there since the project states are in chronological order. I approve!

Campanha:
Do we have guarantee that users will not use IE version below 14? As @leirasf notice that feature of default functions and values are not supported by that browser version.

leirasf:
Here the [Default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)

tomixii:
Users are told to use updated version of Mozilla Firefox or Google Chrome and it looks like they support default parameters.

skosonen:
Chrome and Firefox are the supported browsers. I think we don't have to support IE or Edge at all.

oappicgi:
we DON't have support for IE. There are already dependencies that break viite/oth with IE. I don't think even our Openlayers version supports IE anymore. We only support Chrome and Firefox.

skosonen:
Let's start using ECMAScript 6 features in our code like Tomi has done here. (You might need to enable that in IDEA.)

PR #160, opened 2018-06-08
Viite 1373 name tool restriction of the year of the start date

pedrosag:


viite-UI/src/view/RoadNamingToolWindow.js
```diff
@@ -97,10 +98,21 @@
...
+        futureDate.setFullYear(futureDate.getFullYear() + yearLimit);
+        return {'fieldDate': fieldDate, 'futureDate': futureDate};
+      };
+
         function toggleSaveButton() {
             $('#saveChangedRoads').prop("disabled",
                 !_.every($('input.form-control[data-fieldname="roadName"],input.form-control[data-fieldname="startDate"]'), function (element) {
-                    return $(element).val() !== '';
+                  if ($(element).attr('data-FieldName') === "startDate") {
+                    var datas = getDataObjects($(element).val());
```
tomixii: "datas" is not a english word. Word data is used as plural. So the right grammar would be "data" instead of "datas".

viite-UI/src/view/RoadNamingToolWindow.js
```diff
@@ -120,6 +132,10 @@
                         $('.form-control[data-roadId=' + originalRoadId + '][data-fieldName=endDate]').val(fieldValue);
                         roadNameCollection.setEndDate(originalRoadId, fieldValue);
                     }
+                  var datas = getDataObjects(fieldValue);
```
tomixii: Same "datas"->"data" here.

PR #161, opened 2018-06-11
Viite 1393 loggin when vvh is not responding

oappicgi:
basically added new try block around execution of VVH query that checks for IO exceptions

also removed duplicated code by creating fetchFeaturesAndLog method which has code that was used in multiple places.

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/vvh/VVHClient.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/vvh/VVHClient.scala
@@ -371,13 +371,13 @@ trait VVHClientOperations {
       val client = HttpClientBuilder.create().build()
       try {
         val response = client.execute(request)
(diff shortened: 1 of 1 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/vvh/VVHClient.scala
```diff
@@ -351,33 +362,55 @@ trait VVHClientOperations {
...
-      val fetchVVHStartTime = System.currentTimeMillis()
       val request = new HttpGet(url)
       val client = HttpClientBuilder.create().build()
-      val response = client.execute(request)
       try {
-        mapFields(parse(StreamInput(response.getEntity.getContent)).values.asInstanceOf[Map[String, Any]], url)
-      } finally {
-        response.close()
+        val response = client.execute(request)
+        if (response.getStatusLine.getStatusCode >= 300) {
```
leirasf: You should do that code inside the try finally block. To have the guarantee that the response is close is something fails on those methods.

PR #162, opened 2018-06-11
Viite 1390 korjaa button selects only one splitted link

Campanha:


viite-UI/src/model/SelectedProjectLink.js
```diff
@@ -21,9 +21,13 @@
     };
 
     var openWithErrorMessage = function (id, errorMessage) {
-      current = projectLinkCollection.getProjectLink([id]);
-      ids = [id];
-      eventbus.trigger('projectLink:errorClicked', get(id), errorMessage);
+      current = projectLinkCollection.getProjectLink(id);
+      ids = id;
+        if (!_.isUndefined(current[0].getData().connectedLinkId) && current.length > 2) {
+            openSplit(ids[0], current.length > 1);
```
hestrompa: Since we are already inside of an "if" condition that validates "current.length > 2", we don't need to make the validation to "current.length > 1", we can just send "true" to OpenSplit method.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -223,13 +223,16 @@ object ProjectValidator {
                                     affectedIds: Seq[Long], coordinates: Seq[ProjectCoordinates],
                                     optionalInformation: Option[String])
 
-  def error(id: Long, validationError: ValidationError)(pl: Seq[ProjectLink]): Option[ValidationErrorDetails] = {
-    val (ids, points) = pl.map(pl => (pl.id, GeometryUtils.midPointGeometry(pl.geometry))).unzip
-    if (ids.nonEmpty)
+  def error(id: Long, validationError: ValidationError, info: String = "N/A")(pl: Seq[ProjectLink]): Option[ValidationErrorDetails] = {
+    val (splitLinks, nonSplitLinks) = pl.partition(_.connectedLinkId.nonEmpty)
```
leirasf: You can use the existing validation on the `ProjectLink IsSplit`

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -223,13 +223,16 @@ object ProjectValidator {
...
+    val connectedSplitLinks = splitLinks.flatMap(p => ProjectDAO.getProjectLinksByConnectedLinkId(Seq(p.connectedLinkId.get, p.linkId)))
```
leirasf: We should avoid the queries inside loops because that cause slowness on the application.

viite-UI/src/model/SelectedProjectLink.js
```diff
@@ -21,9 +21,13 @@
     };
 
     var openWithErrorMessage = function (id, errorMessage) {
```
leirasf: The paramenter should be ids instead of id, because now is a array.

PR #165, opened 2018-06-15
VIITE-1399 reservation of road parts fails

tomixii:
Viite now fetches also geometry information of complementary road links.


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
```diff
@@ -140,7 +140,9 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
     if (linkIds.nonEmpty) {
       if (frozenTimeVVHAPIServiceEnabled) {
         vvhClient.frozenTimeRoadLinkData.fetchByLinkIds(linkIds)
-      } else vvhClient.roadLinkData.fetchByLinkIds(linkIds)
+      } else {
+        vvhClient.roadLinkData.fetchByLinkIds(linkIds) ++ vvhClient.complementaryData.fetchByLinkIds(linkIds)
```
leirasf: You can do both with async calls using `fetchByLinkIdsF` and then wait. We also have this method `fetchVVHRoadLinksAndComplementaryFromVVH` improved to be async and use it here.

PR #169, opened 2018-06-19
VIITE-1402 Adding project markers to interactions and refactoring of markers

hestrompa:
Adding markers of project link layer to single and double click interactions.

Refactoring of LinkPropertyLayer markers and ProjectLinkLayer markers.
Refactoring on roadLinkData and projectLinkData handling. Now is always called linkData and we don't need to validate it by name, since they are from different layers

### DIFF
--- a/viite-UI/src/application.js
+++ b/viite-UI/src/application.js
@@ -74,14 +74,14 @@
       })
     });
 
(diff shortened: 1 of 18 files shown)


pedrosag:
Pull request #173 is to be merged onto this branch.

PR #170, opened 2018-06-19
Viite 1398 sending to tr status for project

pedrosag:
Tests on jenkins are passing, check build 1012.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -48,11 +49,17 @@ object ProjectState {
   case object Deleted extends ProjectState {def value = 7; def description = "Poistettu projekti"}
 
   case object ErrorInViite extends ProjectState {
-    def value = 8
+    def value = 8;
```
tomixii: No need for semicolons in Scala.

pedrosag: God dammit... I blame switching too often :)

PR #173, opened 2018-06-19
Viite 1404 changes for zooming and link selecting

pedrosag:


viite-UI/src/view/MapView.js
```diff
@@ -139,12 +139,12 @@
...
 
     $('body').on('keydown', function(evt){
-      if(evt.shiftKey && enableShiftModifier)
+      if (evt.ctrlKey && enableCtrlModifier)
         map.getViewport().style.cursor = "copy";
     });
 
     $('body').on('keyup', function(evt){
-      if(evt.which === 16) // shift key up
+      if (evt.which === 17) // ctrl key up
```
tomixii: I'm wondering if there is keyCode value or something for ctrl key so we wouldn't have to use magic number 17 here.

tomixii: Something like evt.ctrlKey could also work.

leirasf: We should have something like evt.metaKey || evt.ctrlKey to support MAC.

PR #183, opened 2018-06-25
Viite 1391 1392 project validator refactoring

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -31,6 +37,14 @@ object ProjectValidator {
     }
   }
 
+  private def connected(pl1: BaseRoadAddress, pl2: BaseRoadAddress) = {
```
leirasf: Move that method to `BaseRoadAddress` can be usefull.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -502,14 +538,19 @@ object ProjectValidator {
       }
     }
 
-    def getLastValidLinkForParts = {
-      seq.filter(r => r.status != LinkStatus.Terminated && r.endAddrMValue == seq.maxBy(_.endAddrMValue).endAddrMValue)
+    val continuityValidations: Seq[Seq[ValidationErrorDetails] =>  Seq[ValidationErrorDetails]] = Seq(
+      checkContinuityBetweenParts,
```
leirasf: Instead of all the methods has as parameters errors they should only have ` (project: RoadAddressProject, projectLinks: Seq[ProjectLink])`. Then those methods don't need to be nested of the `validateProject` method.

Campanha:
all tests pass

PR #192, opened 2018-06-28
Viite 1401 error in removed end of road parts

Campanha:
For each road number:
**1.** Pick non terminated road parts
**2.** Get bigger part in **1.**
**3.** Get parts that are in **1.**, have EndOfRoad and are lower in number than one in **2.**
**4.** Result will be the ones with error (having endOfRoad, are not terminated, and are not the biggest part for that road Number)


pedrosag:
Merge result successful, tests on jenkins are passing check build 1064.

PR #199, opened 2018-07-02
Viite 1433 prioritize unchanged ordering errors in project validator

Campanha:


Campanha:
all tests passing

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -237,6 +237,42 @@ object ProjectValidator {
...
+        checkForInvalidUnchangedLinks
+      )
+
+      val errors :Seq[ValidationErrorDetails] = actionsOrdering.foldLeft(Seq.empty[ValidationErrorDetails]) { case (errors, validation) =>
+        validation(project, projectLinks) ++ errors
+      }
+      errors.distinct
+  }
+
+  def ProjectLinksValidation(project: RoadAddressProject, projectLinks: Seq[ProjectLink]): Seq[ValidationErrorDetails] =  {
```
tomixii: ActionsOrderingValidation and ProjectLinksValidation could be combined together as a one function since they have a lot of same rows. For example have only one function and give it sequence of functions as a parameter and that sequence can be the "actionsOrdering"/"projectValidations" value. This way may not be the best way to do it but it's one way and you can think about it, change it or don't.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -237,6 +237,42 @@ object ProjectValidator {
...
+  def validateProject(project: RoadAddressProject, projectLinks: Seq[ProjectLink]) :Seq[ValidationErrorDetails] = {
+    time(logger, "Validating project") {
+          ActionsOrderingValidation(project, projectLinks) match {
+            case e if e.nonEmpty => e
+            case _ => ProjectLinksValidation(project, projectLinks)
+          }
+    }
+  }
+
+  def ActionsOrderingValidation(project: RoadAddressProject, projectLinks: Seq[ProjectLink]): Seq[ValidationErrorDetails] = {
```
tomixii: Functions start with lower case. Same thing also in the other function.

PR #218, opened 2018-07-13
VIITE-1457 Fixed connectedDiscontinuousLink validation

tomixii:
Also refactored MinorDiscontinuity validation.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -627,11 +627,31 @@ object ProjectValidator {
     }
 
     def checkContinuityBetweenLinksOnParts: Seq[ValidationErrorDetails] = {
```
oappicgi: It would be good to have test for this method

PR #220, opened 2018-07-16
Viite 1447 missing terminated row from change table

pedrosag:
Warning, this currently does not solve the issues with in the Terminated links, because it is missing the partitioner calculations for them (Task Viite-1448).


pedrosag:
So integration with Viite-1448 is already done due to the use of strategies employed by @leirasf .
All tests are passing, from the UI side and backend side (1240 jenkins build).


I still require to do some MANUAL testing on this but initial results seem good.

pedrosag:
Manual testing complete, results look good, I even attached a picture on the task itself.

PR #228, opened 2018-07-20
Viite 1468 and 1469 floating fixe

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressMapper.scala
```diff
@@ -34,7 +34,7 @@ trait RoadAddressMapper {
       }
     }
 
-    roadAddressMapping.filter(_.matches(ra)).map(m => adjust(m, ra.startMValue, ra.endMValue)).map(adjMap => {
+    roadAddressMapping.filter(_.matches(ra)).map(/*m => adjust(m, ra.startMValue, ra.endMValue)).map(*/adjMap => {
```
pedrosag: Remove the commented code, if it is not needed please..

PR #230, opened 2018-07-25
Viite 1475 change logic of lrm position checker batch

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1437,6 +1437,24 @@ object RoadAddressDAO {
     sqlu"""LOCK TABLE road_address IN SHARE MODE""".execute
   }
 
+  def getAllRoadAddress: Seq[RoadAddress] = {
+    time(logger, "Get road addresses by filter") {
```
hestrompa: Change the comment of timer. If the method gets all road addresses, the filter shouldn't say something "by filter"

PR #232, opened 2018-07-26
Viite 1308 property boundaries on the map

oappicgi:
added WMTS layer for property borders and changed old WMTS API that will expire 2019


conf/dev/digiroad2.properties
```diff
@@ -12,7 +12,7 @@ digiroad2.authenticationTestMode=true
 digiroad2.VKMUrl=http://localhost:8997
 digiroad2.cache.directory=/tmp/digiroad.cache
 digiroad2.viite.importTimeStamp=1514851200000
-http.proxySet=false
+http.proxySet=true
 http.proxyHost=172.17.208.16
 http.proxyPort=8085
-http.nonProxyHosts=172.17.*|localhost|127.*|karttamoottori.maanmittauslaitos.fi
```
skosonen: Is this okay, that these "172.17.*|localhost|127.*|" are removed now? Why not just changing the "karttamoottori.maanmittauslaitos.fi" -part?

oappicgi: at least for me it worked when proxy was set to false.  oag.liikennevirasto.fi should work as long as you are in FTA network, and I couldn't come up with scenario when that wouldn't apply to developers

skosonen: I think it's safer to keep "172.17.\*|localhost|127.\*|" in nonProxyHosts and then these changes should be added to other profiles than dev too.

PR #238, opened 2018-07-31
Viite 1477 adding a floating reason

hestrompa:


skosonen:
When I run the tests on Jenkins, there were 33 failed tests. Check build: #1323
(Hopefully I resolved conflicts correctly when I merged master)

Campanha:
i guess the 33 failing tests are just the missing FLOATING_REASON collumn in the querys

skosonen:
Why we have a separate floating reason column when we could have just the floating-column that has these floating reason values in it? If floating=0, it's not floating, and if it has any other value, it is floating and we would have there the reason as well.

hestrompa:
We discussed about it, and we thought that would be better to keep the current floating logic as it is now, because it is used as a boolean everywhere in the code, and a change like that would require to reset data again and everything related with road address table

skosonen:
I would like to keep that database design clean. Now we would be saying the same thing with two columns.

Code could be changed so that RoadAddress.floating would be of type FloatingReason (or simply Floating).

If we would take this in production now, the current floating road addresses would have anyway the wrong floating reason so I guess it would not be bad if the existing floating ones would have value 1 even though the reason for the floating would be actually something else.

digiroad2-oracle/src/main/resources/db/migration/V0_15__add_floating_reason_column.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "ROAD_ADDRESS"
+  ADD (
+    "FLOATING_REASON" NUMBER(5,0)
```
oappicgi: To my understanding it is bad practice to put column name inside brackets because in some cases it makes column name case sensitive.

oappicgi:
I think Sami makes good point. It is better to do extra work on code side to have less redundant data in database

digiroad2-oracle/src/main/resources/db/migration/V0_15__add_floating_reason_column.sql
skosonen: I'm removing this column.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1602,7 +1603,10 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     }
 
     split.flatMap(pl => {
-      val floatingValue = roadAddress.validTo.isDefined
+
+      // TODO If valid to is defined, what should be the floating reason? - Sami
```
skosonen: What should be the floating reason in this case?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1771,12 +1775,12 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     val floatingValue = source.isDefined && source.get.validTo.isDefined && source.get.validTo.get.isBeforeNow
     val roadAddress = RoadAddress(source.map(_.id).getOrElse(NewRoadAddress), pl.roadNumber, pl.roadPartNumber, pl.roadType, pl.track, pl.discontinuity,
       pl.startAddrMValue, pl.endAddrMValue, None, None, pl.createdBy, pl.linkId, pl.startMValue, pl.endMValue, pl.sideCode,
-      pl.linkGeometryTimeStamp, pl.toCalibrationPoints(), floating = false, geom, pl.linkGeomSource, pl.ely, terminated = NoTermination, source.map(_.commonHistoryId).getOrElse(0))
+      pl.linkGeometryTimeStamp, pl.toCalibrationPoints(), floating = NoFloating, geom, pl.linkGeomSource, pl.ely, terminated = NoTermination, source.map(_.commonHistoryId).getOrElse(0))
     pl.status match {
       case UnChanged =>
-        roadAddress.copy(startDate = source.get.startDate, endDate = source.get.endDate, floating = floatingValue)
+        roadAddress.copy(startDate = source.get.startDate, endDate = source.get.endDate, floating = if (floatingValue) FloatingReason.GeometryChanged else NoFloating) // TODO Floating reason? - Sami
```
skosonen: What should be the floating reason in this case?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1771,12 +1775,12 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
       case Transfer | Numbering =>
-        roadAddress.copy(startDate = Some(project.startDate), floating = floatingValue)
+        roadAddress.copy(startDate = Some(project.startDate), floating = if (floatingValue) FloatingReason.NewAddressGiven else NoFloating) // TODO Floating reason? - Sami
```
skosonen: What should be the floating reason in this case?

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/RoadAddressServiceSpec.scala
```diff
@@ -310,14 +311,14 @@ class RoadAddressServiceSpec extends FunSuite with Matchers{
...
-      address.floating should be (false)
+      address.isFloating should be (false)
       address.geometry shouldNot be (roadLink.geometry)
       roadAddressService.checkRoadAddressFloatingWithoutTX(Set(address.id))
       dynamicSession.rollback()
       val addressUpdated = RoadAddressDAO.queryById(Set(address.id)).head
       addressUpdated.geometry shouldNot be (address.geometry)
       addressUpdated.geometry should be(roadLink.geometry)
-      addressUpdated.floating should be (false)
+      addressUpdated.isFloating should be (true)
```
skosonen: Is this okay that I changed the test and assumed that the implementation was correct?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
Campanha: I think the reason being NewAddressGiven is not that bad.
Floating through:
- ApplyChanges batch -> reason: ApplyChanges
- Find Floating batch -> reason: GeometryChanged or NewAddressGiven, or GapInGeometry 
- Manualy by user in UI -> reason: ManualFloating 

So i would say that we could use that NewAddressGiven since in fact we kind of create one new address road with spliting tool. Or do you think it would be better to have one new type to know were it was done through SplitingTool?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
Campanha: I think the reason being NewAddressGiven is not that bad.
Floating through:

ApplyChanges batch -> reason: ApplyChanges
Find Floating batch -> reason: GeometryChanged or NewAddressGiven, or GapInGeometry
Manualy by user in UI -> reason: ManualFloating
So i would say that we could use that NewAddressGiven since in fact we kind of create one new address road for each side of the cutting point (= 2 new roadaddresses) with spliting tool. Or do you think it would be better to have one new type to know were it was done through SplitingTool?

skosonen:
Tests pass on Jenkins. Check build #1347.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/RoadAddressServiceSpec.scala
Campanha: yep i guess somehow the vvh changed for this link making it floating? At least the changes didnt deal with some new logic to detect floatings. Is it possible that in future the test fails again due to diff vvh geometry for this link?

skosonen: If this is using real VVH, then yes, test can fail in the future. We should change all the tests to use mock VVH so that we can run the tests independently.

skosonen: But not doing this mock VVH change in this pull request.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
skosonen: Maybe I should add a new floating reason: ProjectToRoadAddress

oappicgi:
Jenkins tests Build #1432  passed

PR #239, opened 2018-08-01
VIITE-1482 Wrong AET value in change table

tomixii:
Reverted part of changes made in VIITE-1354 which caused this bug. I guess the VIITE-1354 bugs are happening now again. Also did some refactoring/reformatting here and there.


skosonen:
Tests pass on Jenkins. Check build #1291.

PR #240, opened 2018-08-02
VIITE-1491 changed coefficient logic in floating transfer

Campanha:
Coefficient should be calculated through MValues difference instead geometryLength since we should not trust floating geometry, only its data

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
@@ -59,24 +59,26 @@ object DefloatMapper extends RoadAddressMapper {
     }
 
     val (orderedSource, orderedTarget) = orderRoadAddressLinks(sources, targets)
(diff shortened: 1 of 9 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
```diff
@@ -95,18 +104,57 @@ object DefloatMapper extends RoadAddressMapper {
...
+        (current, setPrecision(applySideCode(mValueLength, mValueLength, current.sideCode)))
+    } else if (mValueLength < mValue) {
+      findStartLinearLocationSource(mValue - mValueLength, links.tail)
+    } else {
+      val dist = applySideCode(mValue, mValueLength, current.sideCode)
+      (current, setPrecision(Math.min(Math.max(0.0, dist), mValueLength)))
+    }
+  }
+
+  private def findStartLinearLocationTarget(mValue: Double, links: Seq[RoadAddressLink]): (RoadAddressLink, Double) = {
```
leirasf: This methed seems to be duplicated.

Campanha:
Tests passing http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/1340/console

PR #241, opened 2018-08-03
Viite 1494 user cannot deselect floatings and the form is not updated

pedrosag:
UI tests are passing and backend tests are passing as well (build 1309).

### DIFF
--- a/viite-UI/src/model/SelectedLinkProperty.js
+++ b/viite-UI/src/model/SelectedLinkProperty.js
@@ -49,8 +49,8 @@
       current = data;
     };
 
-    var idTestForOpen = function(id){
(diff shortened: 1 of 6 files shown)


viite-UI/src/model/SelectedLinkProperty.js
```diff
@@ -50,6 +50,10 @@
       current = data;
     };
 
+    var idTestForOpen = function (id) {
```
leirasf: This name could be something like "canOpenById"

leirasf: I think that what you want to check here is if you have a valid id on the road address could be:
`return !_.isUndefined(id) && id > 0;`
And the name can be `hasValidId`

PR #242, opened 2018-08-03
Viite 1473 modified by field

oappicgi:
added modified_by field and made changed auto_merged to modified_by field so it does not replace original roadaddress creator


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1343,12 +1342,12 @@ object RoadAddressDAO {
     Q.updateNA(query).first
   }
 
-  def create(roadAddresses: Iterable[RoadAddress], createdBy : Option[String] = None): Seq[Long] = {
+  def create(roadAddresses: Iterable[RoadAddress], createdBy : Option[String] = None, modifiedBy : Option[String] = None): Seq[Long] = {
```
skosonen: I guess here the creator and modifier are the same right? If so, then this modifiedBy parameter is unnecessary, just use the value from createdBy.

PR #243, opened 2018-08-06
Viite 1474 error in recalculation for road address history for next non floating link

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -896,6 +896,30 @@ object RoadAddressDAO {
...
 
+  def updateConcerningHistory(roadAddress: RoadAddress, geometry: Option[Seq[Point]] = Option.empty[Seq[Point]]) : Unit = {
+    if (geometry.isDefined) {
+      val startTS = toTimeStamp(roadAddress.startDate)
+      val endTS = toTimeStamp(roadAddress.endDate)
+      val first = geometry.get.head
+      val last = geometry.get.last
+      val (x1, y1, z1, x2, y2, z2) = (first.x, first.y, first.z, last.x, last.y, last.z)
+      val length = GeometryUtils.geometryLength(geometry.get)
+      sqlu"""UPDATE ROAD_ADDRESS
```
skosonen: Why do we have to update the road_number, road_part_number, track_code and discontinuity here? Also wondering about the start and end_dates.

oappicgi: i would also like to know this before approving

pedrosag: I used the usual RoadAddressDAO.update method as a basis, i reckon that if it would update theses fields when invoked, then I assumed this one should update the same fields as well.

skosonen: @leirasf what do you think about this? Should we update here all these fields or just the ones we expect to change here?

leirasf: @skosonen Yes i don't agree with this approach, and beside that this is not correct, because we are changing here the address measures of the history with the value of the current, this not correct.

PR #246, opened 2018-08-09
Viite 1502 find adjacent unknown links sometimes does not return adjacents

pedrosag:


pedrosag:
Tests are passing, jenkins build 1394.

PR #247, opened 2018-08-09
Viite 1501 wrong growing direction after transfer floating

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
```diff
@@ -15,7 +15,10 @@ object DefloatMapper extends RoadAddressMapper {
                     startTargetLink: RoadAddressLink, startTargetM: Double,
                     endTargetLink: RoadAddressLink, endTargetM: Double): RoadAddressMapping = {
       if (startSourceM > endSourceM)
-        formMapping(startSourceLink, endSourceM, endSourceLink, startSourceM,
+        formMapping(startSourceLink, startSourceM, endSourceLink, endSourceM,
+          startTargetLink, endTargetM, endTargetLink, startTargetM)
+      else if (startTargetM > endTargetM)
```
tomixii: I think this should be refactored as
`if (startSourceM > endSourceM || startTargetM > endTargetM)
  formMapping(...)
else
  RoadAddressMapping(...)`

oappicgi: after this  i think it is ready

Campanha: Nice Eagle eye! Actually i think i got some typo in there because the formMapping should not be equal.  If first clause then should be something like "[...]formMapping(startSourceLink, , endSourceM, endSourceLink, startSourceM,..."

I will check and change that.

leirasf:
I will add to that branch the fix of 1503

Campanha:
This has been fixed by and in @leirasf  branch VIITE-1503_Address_gap_on_transfer

PR #254, opened 2018-08-21
Viite 1505 creating two road address segments at the same place

pedrosag:
All tests are passing.
Both UI and on jenkins (build 1387).


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -777,10 +792,25 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
+    val builtMissing = missingLinks.map(ml => RoadAddressLinkBuilder.build(connectedLinks(ml.linkId), ml))
+    val remainingAddresses = roadAddresses.filterNot(ra => builtMissing.exists(_.linkId == ra.linkId))
+
+    val filteredAddresses = remainingAddresses.filter(ra => {
+      !GeometryUtils.withinTolerance(ra.geometry, connectedLinks(ra.linkId).geometry, MaxDistanceDiffAllowed)
+    }).map(fa => {
+      RoadAddressLinkBuilder.build(connectedLinks(fa.linkId), fa, floating = false, Some(connectedLinks(fa.linkId).geometry))
+    })
+
+    builtMissing ++ filteredAddresses.filter(_.floating)
```
tomixii: Doesn't .filter(_.floating) include all and thus is unnecessary since we are setting "floating = false" for all the links?

viite-UI/src/utils/backend-utils.js
```diff
@@ -556,6 +562,14 @@
...
     };
 
+    this.withGetRoadLinkById = function (returnData) {
+      self.getRoadLinkById = function (id, callback) {
+        callback(returnData);
+        return returnData;
+      };
+      return self;
+    };
+
```
tomixii: Idea told me that this function is never used. Is it meant to be used somewhere?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
pedrosag: Not necessarily, the floating flag inside the builder only affects the outcoming RoadLinkType (used by the UI to define the styles, -1 if I recall correctly).
Actual floating value is still defined by the road address in question.

viite-UI/src/utils/backend-utils.js
pedrosag: You are right, it was a remnant of a previous try, deleting it.

Thank you for pointing that out ;)

pedrosag:
Tests still passing, jenkins 1395.

PR #257, opened 2018-08-23
Viite 1507 2 segments on top of each other + wrong growing direction when made not floating

pedrosag:
WARNING: After Sirra and before Tallenna the mAddress values DO NOT add up but after Tallenna the recalculate fixes all the values again.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressMapper.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressMapper.scala
@@ -184,7 +184,7 @@ trait RoadAddressMapper {
     * @param geom2 Geometry two
     */
   def isDirectionMatch(geom1: Seq[Point], geom2: Seq[Point]): Boolean = {
-    val x = distancesBetweenEndPoints(geom1, geom2)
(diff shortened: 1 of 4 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadAddressMapper.scala
```diff
@@ -205,6 +205,23 @@ trait RoadAddressMapper {
...
+  /**
+    * Measure summed distance between two geometries: head-to-head + tail-to-head vs. head-to-tail + tail-to-head
+    * The measurement is taken after the geometries are reduced to the origin point.
+    * @param geom1 Geometry 1
+    * @param geom2 Goemetry 2
+    * @return h2h distance, h2t distance sums
+    */
+
+  def distancesBetweenEndPointsInOrigin(geom1: Seq[Point], geom2: Seq[Point]): (Double, Double) = {
+    val sourcePoint = geom1.minBy(p => p.distance2DTo(Point(0,0,0)))
```
leirasf: You can create a method to transfer the geometry to start at some point, at geometry utils.
`transferGeometry(geom: Seq[Point], to: Point)`

PR #259, opened 2018-08-24
VIITE-1519 Disabled other operations than the new.

skosonen:


oappicgi:
Looks good, To my understanding this will make splitting kinda useless though, but i guess that is what they want (if ennallaan is not allowed)

skosonen:
Whole splitting will be disabled at the beginning so I'm not touching that here.

PR #264, opened 2018-08-28
VIITE-1521 No update on road addresses

skosonen:


skosonen:
One test fails still:
```
[info] - Test change info on links 5622931, 5622953, 499914628 and 499914643 (will refuse transfer) *** FAILED ***
[info]   Set(6163918, 6163919, 6163915, 6163917) was not equal to Set(6163915, 6163919, 6163916, 6163917, 6163918) (RoadAddressServiceSpec.scala:841)
```

skosonen:
This pull-request is not urgent since the next initial import will be done at the end of September and MTK-updates won't be started immediately after that. Also the possible new data structure might render this unnecessary.

PR #268, opened 2018-08-29
Viite 1515 link can not be selected and it is unseen when zooming

pedrosag:


digiroad2-oracle/src/main/resources/db/migration/V0_16__recreate_project_link_history.sql
```diff
@@ -0,0 +1,92 @@
...
+	"REVERSED" NUMBER(1,0) DEFAULT 0,
+	"GEOMETRY" VARCHAR2(4000 BYTE) DEFAULT NULL,
+	"SIDE_CODE" NUMBER(5,0),
+  "START_MEASURE" NUMBER(8,3),
+  "END_MEASURE" NUMBER(8,3),
+  "LINK_ID" NUMBER(38,0),
+  "ADJUSTED_TIMESTAMP" NUMBER(38,0) DEFAULT 0 NOT NULL ENABLE,
+  "LINK_SOURCE" NUMBER(*,0) DEFAULT (1),
+	"CALIBRATION_POINTS_SOURCE" NUMBER DEFAULT 0
+) SEGMENT CREATION IMMEDIATE
```
skosonen: These default values could be removed from here.

PR #270, opened 2018-08-30
Viite common history ids fixes

hestrompa:
This PR contains the tasks:
VIITE-1170 Common history in road address project: terminating road links
VIITE-1171 Common history in road address project: unchanged road links
VIITE-1524 Extra rows, when using the Terminate function

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
@@ -1,15 +1,11 @@
 package fi.liikennevirasto.viite.process
 
 
-import fi.liikennevirasto.digiroad2.GeometryUtils
-import fi.liikennevirasto.digiroad2.asset.SideCode
-import fi.liikennevirasto.digiroad2.asset.SideCode.{AgainstDigitizing, TowardsDigitizing}
 import fi.liikennevirasto.digiroad2.dao.Sequences
 import fi.liikennevirasto.digiroad2.util.Track
 import fi.liikennevirasto.digiroad2.util.Track.Combined
 import fi.liikennevirasto.viite.NewRoadAddress
-import fi.liikennevirasto.viite.dao.{CalibrationPoint, LinkStatus, ProjectLink, RoadAddress}
-import fi.liikennevirasto.viite.process.ProjectSectionCalculator.getClass
+import fi.liikennevirasto.viite.dao.{LinkStatus, ProjectLink, RoadAddress}
 import fi.liikennevirasto.viite.util.CalibrationPointsUtils.fillCPs
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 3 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -46,6 +42,34 @@ object CommonHistoryFiller {
     }.toSeq ++ newRoadAddresses.filterNot(ra => unchangedLinks.map(_.roadAddressId).contains(ra.id))
   }
 
+  private def applyTerminated(currentRoadAddresses: Seq[RoadAddress])(projectLinks: Seq[ProjectLink], newRoadAddresses: Seq[RoadAddress]): Seq[RoadAddress] = {
+    val terminatedLinks = projectLinks.filter(_.status == LinkStatus.Terminated)
+
+    terminatedLinks.groupBy(pl => (pl.roadNumber, pl.roadPartNumber, pl.track, pl.roadType)).flatMap {
+      case ((roadNumber, roadPartNumber, trackCode, roadType), groupedLinks) =>
+        val roadAddressesToReturn = newRoadAddresses.filter(ra => groupedLinks.sortBy(_.startAddrMValue).map(_.roadAddressId).contains(ra.id))
```
tomixii: Why groupedLinks needs to be sorted here? Also .map(_.roadAddressId).contains(ra.id) loops through the array twice. It can be refactored like this: .exists(_.roadAddressId == ra.id). This will loop through the array only once and thus the time used for this operation is reduced to half.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -46,6 +42,34 @@ object CommonHistoryFiller {
...
+    terminatedLinks.groupBy(pl => (pl.roadNumber, pl.roadPartNumber, pl.track, pl.roadType)).flatMap {
+      case ((roadNumber, roadPartNumber, trackCode, roadType), groupedLinks) =>
+        val roadAddressesToReturn = newRoadAddresses.filter(ra => groupedLinks.sortBy(_.startAddrMValue).map(_.roadAddressId).contains(ra.id))
+        val roadAddressesToCompare = currentRoadAddresses.filter(ra => ra.roadNumber == roadNumber && ra.roadPartNumber == roadPartNumber && ra.track == trackCode && ra.roadType == roadType).sortBy(_.startAddrMValue)
+        if (groupedLinks.nonEmpty && roadAddressesToCompare.nonEmpty && groupedLinks.lengthCompare(roadAddressesToCompare.length) == 0 && (groupedLinks.last.endAddrMValue - groupedLinks.head.startAddrMValue == roadAddressesToCompare.last.endAddrMValue - roadAddressesToCompare.head.startAddrMValue)) {
+          roadAddressesToReturn
+        } else {
+          roadAddressesToReturn.groupBy(_.commonHistoryId).values.toSeq.foldLeft(Seq.empty[RoadAddress]) {
+            case (seqComplete, addresses) =>
+              val currentsInsideGroup = roadAddressesToCompare.filter(ra => addresses.map(_.id).contains(ra.id))
```
tomixii: `addresses.map(_.id).contains(ra.id)` can be refactored here too.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
```diff
@@ -46,6 +42,34 @@ object CommonHistoryFiller {
...
+              } else {
+                val changedAddresses = addresses.sortBy(_.startAddrMValue).foldLeft(Seq.empty[RoadAddress]) {
+                  case (seq, address) =>
+                    seq ++ Seq(fillCommonHistoryId(seq, address))
+                }
+                seqComplete ++ changedAddresses
+              }
+          }
+        }
+    }.toSeq ++ newRoadAddresses.filterNot(ra => terminatedLinks.map(_.roadAddressId).contains(ra.id))
```
tomixii: Refacotring can be done here too.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/CommonHistoryFiller.scala
hestrompa: We need to sort the links because we validate the length of the segment on this method, and it is easier if they are sorted. But you have a point on the loops, I will change those

tomixii: groupedLinks array doesn't keep the changes what sortBy does here. It creates kind of new array when sorted and then calls map to that array. groupedLinks is still unsorted. Might be that it is sorted already in the beginning and that's why this code works, but that sortBy doesn't affect to the rest of the method at all.

tomixii:
I didn't mean that all sortBy function should be removed. Just the ones which are inside lambda functions which doesn't accomplish anything :)

PR #272, opened 2018-08-31
Viite 1526 selection corrections

pedrosag:


Campanha:
Regarding the "Removed a un-needed filter on the front end that was removing wrong links."
The unknown adjacents are now appearing which is fine. But it also appears to be duplicated, at least for the test i did, along with duplicated markers in map (for e.g. B drawn in top of A)

skosonen:
Henni also noticed these duplicated markers drawn on top of each others on devtest. So I think that issue was introduced already on some earlier feature.

viite-UI/src/model/SelectedLinkProperty.js
```diff
@@ -339,11 +339,9 @@
                 return _.some(foundDuplicatedLink);
             });
             previousAdjacents = filteredAdjacents;
-            var unknownAdjacents = _.reject(filteredAdjacents, function(t){
-               return t.roadLinkType === RoadLinkType.FloatingRoadLinkType.value;
-            });
+
             var markedRoads = {
-              "adjacents": _.map(applicationModel.selectionTypeIs(selectionType.Floating) ? unknownAdjacents : unknownAdjacents, function (a, index) {
+              "adjacents": _.map(applicationModel.selectionTypeIs(selectionType.Floating) ? filteredAdjacents : filteredAdjacents, function (a, index) {
```
tomixii: Why there is conditional (ternary) operator here? It same as 
`if(something)
  value
else
  value`
which doesn't make any sense since it chooses filteredAdjacents anyway. Or am I missing something?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -757,25 +757,34 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
+          buildFloatingRoadAddressLink(fh, floatings.filter(_.linkId == fh.linkId))
+        })
       }
-      if (historyLinks.nonEmpty) {
-        val historyLinkAddresses = time(logger, "Build history link addresses") {
-          historyLinks.flatMap(fh => {
-            buildFloatingRoadAddressLink(fh, floatings.filter(_.linkId == fh.linkId))
-          })
-        }
-        val selected = historyLinkAddresses.find(_.id == id).getOrElse(historyLinkAddresses.find(_.linkId == linkId).get)
```
tomixii: Replacing `get` in the end with `getOrElse(Option.empty[RoadAddressLink])` would do the same thing as the `if-else if-else`, right?

pedrosag:
Tests are passing, check build http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/1454/console.

UI tests are passing as well:
image

PR #274, opened 2018-09-06
VIITE-1170 fixing end dates for history links.

hestrompa:
Now the end date used is the project date instead of the current date.

Unit test added


hestrompa:
Tests are running ok

PR #275, opened 2018-09-07
Viite 1502 fix unknown adjacency problems

Campanha:


Campanha:
All tests passing http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/1444/console

PR #280, opened 2018-09-11
VIITE-1171 - Adding history to unchanged addresses 

hestrompa:
Saving history links when we have changes in addresses and using the unchanged operation in project


pedrosag:
Tests are passing:
http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/1466/console
UI Tests are also passing.

PR #282, opened 2018-09-12
VIITE-1568: Multiple road names are not saving correctly

hestrompa:


hestrompa:
This will be merged on a new branch created from the last deployed version to production

PR #286, opened 2018-09-13
VIITE-1561 added checks to fix the validations

hestrompa:
Previous Tomi PR: 
Fixed road address find so that right road(s) are found. Now discontinuity End of Road is right for ELY border when the same road (same road number) doesn't continue in the next ELY. I'm not sure if the gruntfile.js was supposed to include in the pull request or not but it's there anyway.


tomixii:
gruntfile is cleaned from this PR.

PR #295, opened 2018-09-19
VIITE-1580 (FAST TRACK) Ctrl singleclick works now like ctrl doubleclick

tomixii:
Users reported that they want ctrl + single click work like ctrl + double click. This makes selecting roadlinks easier. I solved this by making some modifications to selections and calling showDoubleClickChanges instead of showSingleClickChanges if ctrl is pressed.


Campanha:
**ctrl singleclick** link its seems not to work like **ctrl doubleclick** when selecting grey links

tomixii:
Oh yea, I forgot to mention that it is meant to work only in project mode as users wanted.

tomixii:
There was one test failing. I fixed it.

PR #302, opened 2018-10-01
VIITE-1599:

hestrompa:
- Hiding suravage links on project link layer
- Disable suravage links toggle button on project link layer.

### DIFF
--- a/viite-UI/src/view/TileMapSelector.js
+++ b/viite-UI/src/view/TileMapSelector.js
@@ -42,13 +42,9 @@
     });
     $('#roadsVisibleCheckbox').change(function () {
       applicationModel.toggleRoadVisibility();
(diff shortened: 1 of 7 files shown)


viite-UI/src/view/TileMapSelector.js
```diff
@@ -42,13 +42,9 @@
     });
     $('#roadsVisibleCheckbox').change(function () {
       applicationModel.toggleRoadVisibility();
-      switch(applicationModel.getSelectedLayer()) {
```
tomixii: If we remove this switch both main layers are set visible or not visible when the checkbox is clicked. Is this what we want? Do we want to show LinkPropertyLayer also in project mode and vice versa?

PR #317, opened 2018-10-09
Viite 1619 m values of the two track roads are not same in the database and in the change table

pedrosag:


pedrosag:
Tests are passing: Jenkins build 1516.

pedrosag:
@skosonen Comment updated to reflect new changes.

PR #324, opened 2018-10-12
Viite 1601 v2 problems with new two track road

pedrosag:
So basically this is an experiment/prof of concept. The main change here are:
A: the initial calculation of the directional vector in the case of having NO other links whose side code is already defined (Not unknown, value = 99).

The endpoints are now defined by the closest target  to a supplied direction vector between the 1st geometry point - last geometry point or the last geometry point - 1st geometry point.

The calculation of the starting point for project recalculation when dealing with 2 tracks (Left, Right). It is now determined by temporarily moving both track geometry to the origin (0.0, 0.0, 0.0) and calculating the distance between the endpoints from/to each track (similar to the logic applied in the defloater tool).

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
@@ -102,7 +102,14 @@ class DefaultSectionCalculatorStrategy extends RoadAddressSectionCalculatorStrat
     else {
       // Get left track non-connected points and find the closest to right track starting point
       val leftLinks = newLinks.filter(_.track != Track.RightSide) ++ oldLinks.filter(_.track != Track.RightSide)
-      val leftPoints = TrackSectionOrder.findOnceConnectedLinks(leftLinks).keys
+      val leftPoints = if (leftLinks.size == 1){
+        leftLinks.flatMap(l => {
+          val (p1, p2) = GeometryUtils.geometryEndpoints(l.geometry)
+          Seq(p1 -> l, p2 -> l)
+        }).groupBy(_._1).mapValues(_.map(_._2).toSeq.distinct).keys
+      } else {
+        TrackSectionOrder.findOnceConnectedLinks(leftLinks).keys
+      }
 
 
       val (d1, d2) = GeometryUtils.distancesBetweenEndPointsInOrigin(rightPoints.toSeq, leftPoints.toSeq)
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
@@ -22,8 +22,9 @@ class DefaultSectionCalculatorStrategy extends RoadAddressSectionCalculatorStrat
       k -> (groupedProjectLinks.getOrElse(k, Seq()), groupedOldLinks.getOrElse(k, Seq())))
     group.flatMap { case (part, (projectLinks, oldLinks)) =>
       try {
-        val (right, left) = TrackSectionOrder.orderProjectLinksTopologyByGeometry(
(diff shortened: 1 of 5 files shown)


viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -731,7 +731,8 @@
       if (map.getView().getZoom() > zoomlevels.minZoomForDirectionalMarkers) {
         var addMarkersToLayer = function(links, layer) {
           var directionMarkers = _.filter(links, function (projectLink) {
-            return projectLink.id !== 0 || (projectLink.id === 0 && projectLink.anomaly === Anomaly.NoAddressGiven.value) || (projectLink.id === 0 && projectLink.roadLinkType === RoadLinkType.FloatingRoadLinkType.value);
+            var acceptedLinks = projectLink.id !== 0 || (projectLink.id === 0 && projectLink.anomaly === Anomaly.NoAddressGiven.value) || (projectLink.id === 0 && projectLink.roadLinkType === RoadLinkType.FloatingRoadLinkType.value);
+            return acceptedLinks && projectLink.sideCode !== 99;
```
Campanha: should use LinkValues.SideCode.Unknown.value instead 99

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
```diff
@@ -123,33 +148,31 @@ class DefaultSectionCalculatorStrategy extends RoadAddressSectionCalculatorStrat
...
-          case _ =>
-            calculateDirectionVector(remainLinks)
+        // Grab all the endpoints of the links
+        val direction = if (remainLinks.exists(_.sideCode != SideCode.Unknown)) {
+          remainLinks.filter(_.sideCode != SideCode.Unknown).map(p => p.endPoint - p.startingPoint).fold(Vector3d(0, 0, 0)) { case (v1, v2) => v1 + v2 }.normalize2D()
+        } else {
+          val sampleLink = remainLinks.head.geometry.head
+          if (remainLinks.flatMap(_.geometry).forall(geo => geo.y == sampleLink.y)) {
+            // Very rare case
+            // ALL Y Values are equal on the tracks check the X Value
```
skosonen: Does this mean that y-values at starting and ending points are the same?

pedrosag: @skosonen Exactly, the very rare case where the whole collective of project links represent a perfectly parallel line to the X axis. In that case there is no variation on the Y values.

skosonen: Line doesn't have to be straight, just if the starting and ending points have the same y-values, then the addressing should grow from west to the east (left to the right).

pedrosag: @skosonen All right, doing that change then.

pedrosag:
@skosonen change done.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
leirasf: This will not work when there is at least one project link with different Y values. Because then the direction vector used for the project link with the same Y value will be also (0.0, 1.0, 0.0)

pedrosag:
@skosonen Sorry I forgot to push the change... it was a local commit until now.

Campanha:
git is having problems. Last commit by @pedrosag due to code review changes https://github.com/finnishtransportagency/viite/commit/b3247a02c903987e24daf6b8e42b7f3d9476b98e

Campanha:
git is having problems. Last commit by @pedrosag due to code review changes https://github.com/finnishtransportagency/viite/commit/b3247a02c903987e24daf6b8e42b7f3d9476b98e

PR #328, opened 2018-10-17
VIITE-1634 listen for ajorata button change

plourenco90:


plourenco90:
This fix is also included on new version of VIITE in branch of VIITE-1540

viite-UI/src/view/ProjectEditForm.js
```diff
@@ -483,6 +483,10 @@
         }
       });
 
+      rootElement.on('change', '#trackCodeDropdown', function () {
```
pedrosag: Quick question @plourenco90 why is checkInputs being called with a ('.project-') parameter when the function accepts no input?

plourenco90: Its because this method is used in the split validation as well, so this is the way it differs from witch form is validating

PR #345, opened 2018-11-02
VIITE-1169 v2 common history in road address project adding a new road link

skosonen:


hestrompa:
This task should be no longer necessary on master after this new logic implementation

PR #346, opened 2018-11-06
Viite 1657 recalculation in two tracks does not adjust addr m values v3

pedrosag:
Tests are passing, jenkins build number: 58

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -240,7 +240,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
           (ProjectDAO.getRoadAddressProjectById(projectId).getOrElse(throw new RuntimeException(s"Missing project $projectId")),
             ProjectDAO.getProjectLinksByProjectRoadPart(roadNumber, roadPartNumber, projectId))
         }
-        val reversed = if(existingProjectLinks.nonEmpty) existingProjectLinks.forall(_.reversed) else false
+        val reversed = if (existingProjectLinks.nonEmpty) existingProjectLinks.forall(_.reversed) else false
         val projectLinks: Seq[ProjectLink] = linkIds.map { id =>
           newProjectLink(roadLinks(id), project, roadNumber, roadPartNumber, track, discontinuity, roadType, roadEly, roadName, reversed)
         }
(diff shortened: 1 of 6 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -203,6 +203,15 @@ case class ProjectLink(id: Long, roadNumber: Long, roadPartNumber: Long, track:
...
     }
   }
+
+  def hasCalibrationPointAt(addressMValue: Long): Boolean = {
+    calibrationPoints match {
+      case (None, None) => false
+      case (Some(cp1), None) => cp1.addressMValue == addressMValue
+      case (None, Some(cp1)) => cp1.addressMValue == addressMValue
+      case (Some(cp1), Some(cp2)) => cp1.addressMValue == addressMValue || cp2.addressMValue == addressMValue
+    }
```
leirasf: You can change the code to be like that:
```
calibrationPoints match {
      case (None, None) => false
      case (Some(cp1), _) => cp1.addressMValue == addressMValue
      case (_, Some(cp1)) => cp1.addressMValue == addressMValue
```

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1202,12 +1203,22 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
-            CalibrationPointDAO.createCalibrationPoint(calibrationPoint)
-          else
-            CalibrationPointDAO.updateSpecificCalibrationPointMeasures(foundCalibrationPoint.head.id, addressM.toDouble - endSegment.startMValue, addressM)
-          Seq(CalibrationPoint)
+          val calibrationPointIsPresent = toUpdateLinks.find(ul => ul.projectId == projectId && ul.startAddrMValue == endSegment.startAddrMValue) match {
+            case Some(projectLink) => {
+              projectLink.hasCalibrationPointAt(calibrationPoint.addressMValue)
+            }
+            case _ => false
+          }
```
leirasf: I think this can be simplified:
`val calibrationPointIsPresent = toUpdateLinks.exists(pl => pl.hasCalibrationPointAt(calibrationPoint.addressMValue))`
Also the variable should be renamed.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
pedrosag: Not in this case because the value we are trying to find can be in either of the calibration points, if we replace by that, the last expression becomes unreachable and we become unable to find the value we seek when we have calibration points in both ends.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
pedrosag: Changed.

pedrosag:
Tests be passing, jenkins build number 71.

PR #352, opened 2018-11-08
VIITE-1373_removed_future_date_since_original_start_date

pedrosag:
Removed the check of the date + 5 years when dealing with the original date on the link, the +5 years should ALWAYS be from the sysdate.
Added the ability to focus the calendar on the original start date.


skosonen:
Product owner decided not to deploy a new version to production with the old structure.

hestrompa:
Closing this PR because we will not need to use VIITE_OLD structure anymore

PR #355, opened 2018-11-09
Viite 1668 url search by linkid in suravage links does not work

Campanha:


viite-UI/src/router.js
```diff
@@ -37,10 +37,8 @@
               models.selectedLinkProperty.open(response.linkId, response.id, true);
               eventbus.trigger('linkProperties:reselect');
             });
-            backend.getRoadAddressMiddlePoint(linkId, function (response) {
-                map.getView().setCenter([response.x, response.y]);
-                map.getView().setZoom(12);
-            });
+              map.getView().setCenter([response.middlePoint.x, response.middlePoint.y]);
+              map.getView().setZoom(12);
```
pedrosag: This zoom level can be included in zoom-levels.js and then used here.

PR #356, opened 2018-11-09
VIITE-1677 - Tienimen päivittyminen automaattisesti tienumeron perusteella / Automatic update of the road name by the road number

pedrosag:
Changed the condition on the frontend.
Made it so that the return data is in a format legible by the frontend.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadNameService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadNameService.scala
@@ -120,7 +120,7 @@ class RoadNameService() {
         if (currentRoadNames.isEmpty) {
           val projectRoadNames = ProjectLinkNameDAO.get(roadNumber, projectID)
           if (projectRoadNames.isEmpty) {
-            return None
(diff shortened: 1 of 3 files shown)


pedrosag:
Tests running ok: Jenkins build 63

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadNameService.scala
```diff
@@ -120,7 +120,7 @@ class RoadNameService() {
         if (currentRoadNames.isEmpty) {
           val projectRoadNames = ProjectLinkNameDAO.get(roadNumber, projectID)
           if (projectRoadNames.isEmpty) {
-            return None
+            Some(Map("roadName" -> None, "isCurrent" -> false))
```
leirasf: Since we don't need to return at any time a `None` we should remove from the `getRoadNameByNumber` method the Option.

pedrosag: Option[] and Somes() removed from the method.

PR #357, opened 2018-11-09
VIITE-1677 - Tienimen päivittyminen automaattisesti tienumeron perusteella / Automatic update of the road name by the road number

pedrosag:
Changed the condition on the frontend.
Made it so that the return data is in a format legible by the frontend.


pedrosag:
Missing a re-evaluation of the tests.

PR #374, opened 2018-11-30
VIITE 1677 and VIITE 1707

pedrosag:
In this PR I added the Viite-1677 and Viite-1707 tasks since they deal with the topic of road names.

Tests are passing, check build 121 in jenkins.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -35,7 +35,35 @@ import scala.concurrent.duration.Duration
...
+
+object RoadNameSource {
+  val values = Set(UnknownSource, ProjectLinkSource, RoadAddressSource)
+
+  def apply (value: Long): RoadNameSource = {
+    values.find(_.value == value).getOrElse(UnknownSource)
+  }
+
+  case object UnknownSource extends RoadNameSource {
+    def value = -1
```
Campanha: hestrompa 4 days ago  Collaborator
Please change this so we don't have negative values.
Add it as 99, or as 0 and the others as 1 and 2

skosonen:
Product owner decided not to deploy a new version to production with the old structure.

hestrompa:
Closing this PR because we will not need to use VIITE_OLD structure anymore

PR #375, opened 2018-12-03
VIITE-1677 VIITE-1707 master road names fixes

pedrosag:
Since master already has the Viite-1677 code on it then it was only a matter of adding the 1707 code.

### DIFF
--- a/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
+++ b/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
@@ -175,6 +175,20 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
           .map(midPoint)
           .getOrElse(Map("success" -> false, "reason" -> ("ID:" + id + " not found")))
     }
+
+  }
+
+  get("/roadlinks/project/prefillfromvvh") {
+    val linkId = params("linkId").toLong
+    val currentProjectId = params("currentProjectId").toLong
+    time(logger, s"GET request for /roadlinks/project/prefillfromvvh (linkId: $linkId, projectId: $currentProjectId)") {
+      projectService.fetchPreFillFromVVH(linkId, currentProjectId) match {
+        case Right(preFillInfo) => {
+          Map("success" -> true, "roadNumber" -> preFillInfo.RoadNumber, "roadPartNumber" -> preFillInfo.RoadPart, "roadName" -> preFillInfo.roadName, "roadNameSource" -> preFillInfo.roadNameSource.value)
+        }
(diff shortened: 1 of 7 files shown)


pedrosag:
All tests are green in build 127.

viite-UI/src/utils/LinkValues.js
```diff
@@ -130,6 +130,12 @@
         SuravageRoadLink:           {value: 4, description: "SuravageRoadLink"}
     };
 
+    root.RoadNameSource = {
+        UnknownSource:              {value: -1, source: "Unknown Source"},
```
Campanha: would be better to change from "source" to "description" like we use in rest of LinkValues file

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -175,6 +175,20 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
           .map(midPoint)
           .getOrElse(Map("success" -> false, "reason" -> ("ID:" + id + " not found")))
     }
+
+  }
+
+  get("/roadlinks/project/prefillfromvvh") {
```
Campanha: duplicate entry point

pedrosag: Already removed it mate, thank you though.

viite-UI/src/utils/LinkValues.js
pedrosag: Done.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -34,7 +34,35 @@ import scala.concurrent.duration.Duration
...
+
+object RoadNameSource {
+  val values = Set(UnknownSource, ProjectLinkSource, RoadAddressSource)
+
+  def apply (value: Long): RoadNameSource = {
+    values.find(_.value == value).getOrElse(UnknownSource)
+  }
+
+  case object UnknownSource extends RoadNameSource {
+    def value = -1
```
hestrompa: Please change this so we don't have negative values.
Add it as 99, or as 0 and the others as 1 and 2

viite-UI/src/utils/LinkValues.js
```diff
@@ -130,6 +130,12 @@
         SuravageRoadLink:           {value: 4, description: "SuravageRoadLink"}
     };
 
+    root.RoadNameSource = {
+        UnknownSource:              {value: -1, description: "Unknown Source"},
```
hestrompa: Same here, use Unknown value as 99 instead of -1

PR #379, opened 2018-12-06
Viite 1239 splitting suravage links

Campanha:
Spliting code was fixed.
Currently the tool, check box, and backend calls for suravage, are disabled


Campanha:
Tests passing, check http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/173/console

PR #380, opened 2018-12-06
VIITE-1721 and VIITE-1051-road name when changing road number & Expire road names when roadway changes have Termination change type

GoAlvesCGI:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1708,47 +1708,118 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
-          logger.error(s" There are no addresses to update, rollbacking update ${project.id}")
-          throw new InvalidAddressDataException(s"There are no addresses to update , rollbacking update ${project.id}")
+    if (projectLinks.isEmpty) {
+      logger.error(s" There are no addresses to update, rollbacking update ${project.id}")
+      throw new InvalidAddressDataException(s"There are no addresses to update , rollbacking update ${project.id}")
+    }
+
+    projectLinkDAO.moveProjectLinksToHistory(projectID)
+    logger.info(s"Moving project links to project link history.")
+    //        handleNewRoadNames(projectLinks, project)
```
Campanha: should commented code be removed?

GoAlvesCGI: Yes. I removed the commented function call to 'handleNewRoadNames' since it is now called inside the try catch block.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1792,7 +1862,9 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   }
 
-  private def getProjectsWaitingForTR: Seq[Long] = {
+  private def getProjectsWaitingForTR: Seq[Long]
```
pedrosag: A bit way too much of spacing.

GoAlvesCGI: auto-formatters... what's the standard in this project?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1733,17 +1735,93 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+            }
+          }
+        }
+      }
+    })
+
+    RoadNameDAO.create(roadNames)
+  }
+
+  def handleTerminatedRoadwayChanges(roadwayChanges: Seq[ProjectRoadwayChange]) = {
```
elenaols: In case, if only the whole road terminated, the road name can be terminated as well. If there are left other road parts of the same road, roan name remains valid

GoAlvesCGI: I'm validating, by roadnumber, if there are any roadways. Only if there are no roadways, the road name is terminated

PR #384, opened 2018-12-11
Viite 1724 information of the link changes when zooming

pedrosag:
Jenkins tests are green, check build 158 on jenkins.


hestrompa:
I'm testing this, and VIITE is too slow after this implementation. Probably we need to think in another solution

skosonen:
(Speling mistake in the commit comment "no failing" -> "now failing")

skosonen:
I found some algorithms for decimating polyline:

http://geomalgorithms.com/a16-_decimate-1.html

PR #393, opened 2018-12-17
VIITE-1724 information of the link changes when zooming

pedrosag:


pedrosag:
All tests are green to go on build 172.

hestrompa:
We will build a job to update geometries instead of changing geometries on import

PR #409, opened 2019-01-07
VIITE-1756 Problems with new 2-track roads and roundabouts

hestrompa:
Fixing on DefaultSectionCalculatorStrategy when trying to find the start point


hestrompa:
Tests are running fine in [#231](http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/231)

PR #410, opened 2019-01-08
Viite 1760 all links should be automatically renumbered

Campanha:


Campanha:
tests pass on build http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/237/console

PR #414, opened 2019-01-09
VIITE-1759 Checking if length has changed by comparing current roadwa…

skosonen:
…y to the change table target roadway.


skosonen:
Tests pass on Jenkins.

PR #419, opened 2019-01-15
VIITE-1716 Added new network error types to road network checker

Campanha:


digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/RoadNetworkServiceSpec.scala
```diff
@@ -24,6 +24,13 @@ class RoadNetworkServiceSpec extends FunSuite with Matchers{
     }
   }
 
+  test("test" ) {
```
Campanha: this code can be removed

GoAlvesCGI: removed

PR #422, opened 2019-01-16
VIITE-1772 Fixed finding of parallel link for the link with discontin…

skosonen:
…uity.


skosonen:
Tests pass on Jenkins: #268

PR #425, opened 2019-01-16
VIITE-1774 Replaced error message when status code >= 500 in TR interface

GoAlvesCGI:


GoAlvesCGI:
@elenaols the trMessageRefusal is used when TR replies with error code >= 500. And trUnreachableMessage is used when TR is down. So I think we should keep both of them, at least for now.

GoAlvesCGI:
Sorry @elenaols I didn't realize that the text of messages were exactly the same. So, I made a new commit, and now we only have a single message, to replace the 2 previous ones.

PR #426, opened 2019-01-17
VIITE-1775 - Käännetty-check boxin ruksi puuttuu kääntö-projektissa /Käännetty - selected check box is missing in reversing project

pedrosag:
Removed the resetting of SideCode from the ProjectSectionCalculator, before calling a specific strategy's assignMValues


pedrosag:
Also all tests are green:

http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/273/console

PR #427, opened 2019-01-17
Viite 1756 - fixing problems with directions in new operation

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadwayDAO.scala
```diff
@@ -238,6 +238,32 @@ trait BaseRoadAddress {
...
+  lazy val endPoint: Point = (sideCode == SideCode.AgainstDigitizing, reversed) match {
+    case (true, true) | (false, false) =>
+      //reversed for both SideCodes
+      geometry.last
+    case (true, false) | (false, true) =>
+      //NOT reversed for both SideCodes
+      geometry.head
+  }
+
+  def getEndPoints: (Point, Point) = {
```
Campanha: You can probably use the existing method getEndPoints for ProjectLink class

hestrompa: No, I removed it from ProjectLinkDAO to this parent class BaseRoadAddress, because we need this getEndPoints for RoadAddress and ProjectLink classes

skosonen:
(ProjectLink.getEndPoints doesn't take in account "reversed"-attribute now. In this case it should not be a problem since this is about new operation, but I'm afraid that this will cause problems in other than new operations.)

PR #429, opened 2019-01-18
VIITE-1772 When the discontinuity is on the left track, …

skosonen:
…the nearest link on the opposite track should be ignored and the parallel link should be searched after that.


skosonen:
Let's wait for client's final approval for this change in algorithm before merging this.

PR #435, opened 2019-01-28
Viite 1724 clean

pedrosag:
All tests are green on jenkins build 323.

### DIFF
--- a/digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
+++ b/digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
@@ -252,14 +252,11 @@ class GeometryUtilsSpec extends FunSuite with Matchers {
 
   test("Test geometryUtils.geometryReduction() When using a very large geometry Then return a smaller geometry") {
     val largeGeom = (0 to 100000 by 10).sliding(2).map(t => {
-      Point(t.head, t.last, 0.0)
+      Point(t.head, t.last)
     }).toSeq
-    largeGeom.head should be(Point(0.0, 10.0, 0.0))
-    largeGeom.last should be(Point(99990.0, 100000.0, 0.0))
-    val begin = System.currentTimeMillis()
+    largeGeom.head should be(Point(0.0, 10.0))
+    largeGeom.last should be(Point(99990.0, 100000.0))
     val reducedGeom = geometryReduction(largeGeom)
-    val duration = System.currentTimeMillis() - begin
-    println(s"GeometryReduction from ${largeGeom.size} done in $duration ms to ${reducedGeom.size}")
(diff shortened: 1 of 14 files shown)


digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
```diff
@@ -249,4 +249,17 @@ class GeometryUtilsSpec extends FunSuite with Matchers {
...
+  test("Test geometryUtils.geometryReduction() When using a very large geometry Then return a smaller geometry") {
+    val largeGeom = (0 to 100000 by 10).sliding(2).map(t => {
+      Point(t.head, t.last, 0.0)
+    }).toSeq
+    largeGeom.head should be(Point(0.0, 10.0, 0.0))
+    largeGeom.last should be(Point(99990.0, 100000.0, 0.0))
+    val begin = System.currentTimeMillis()
+    val reducedGeom = geometryReduction(largeGeom)
+    val duration = System.currentTimeMillis() - begin
+    println(s"GeometryReduction from ${largeGeom.size} done in $duration ms to ${reducedGeom.size}")
```
skosonen: Is this okay that we print out something in Unit test?

digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
```diff
@@ -249,4 +249,17 @@ class GeometryUtilsSpec extends FunSuite with Matchers {
...
+    reducedGeom.size < largeGeom.size should be (true)
```
skosonen: Maybe here would be good to test that the reducedGeom.size is something expected, not just smaller than the original.

digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
pedrosag: @skosonen I checked with @hestrompa and he said it was ok...

digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
pedrosag: @skosonen I also added now the checking that the start and end points of both geometries are the same and that the reduced geometry is a subset of the original one.

digiroad2-geo/src/test/scala/fi/liikennevirasto/digiroad2/GeometryUtilsSpec.scala
hestrompa: I think that we can use log instead of print

PR #438, opened 2019-01-29
VIITE-1775 Fixed missing reversed roads in reversing project

GoAlvesCGI:


GoAlvesCGI:
Only merge when https://extranet.vayla.fi/jira/browse/VIITE-1783 is ready

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/TrackSectionOrder.scala
```diff
@@ -288,10 +288,10 @@ object TrackSectionOrder {
             (getOppositeEnd(l, currentPoint), l)
         }
         // Check if link direction needs to be turned and choose next point
-        val sideCode = (nextLink.geometry.last == nextPoint, nextLink.reversed && (unprocessed ++ ready).size == 1) match {
-          case (false, false) | (true, true) =>
+        val sideCode = (nextLink.geometry.last == nextPoint, nextLink.reversed && ready.isEmpty) match {
+          case (false, _) | (true, true) =>
```
Campanha: we could probably have done something like
case (true, false) =>SideCode.TowardsDigitizing	            
case (_) => SideCode.TowardsDigitizing

GoAlvesCGI: done :)

PR #454, opened 2019-02-11
Viite 87 initial check fix

elenaols:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/RoadNetworkChecker.scala
```diff
@@ -31,19 +31,24 @@ class RoadNetworkChecker(roadLinkService: RoadLinkService) {
       } else {
         val roadNumbers = roadwayDAO.getValidRoadNumbers
         val chunks = generateChunks(roadNumbers, 500)
+        roadNetworkDAO.getLatestRoadNetworkVersionId.getOrElse(
+          roadNetworkDAO.createPublishedRoadNetwork(Sequences.nextPublishedRoadNetworkId)
```
skosonen: Why do we create published road network two times? Here and at the row 47?

skosonen: If we have the initial situation when we don't have yet any published road networks, do we need to create before checking one published road network first?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadNetworkDAO.scala
```diff
@@ -16,8 +16,8 @@ class RoadNetworkDAO {
 
   protected def logger = LoggerFactory.getLogger(getClass)
 
-  def createPublishedRoadNetwork: Unit = {
-    sqlu"""INSERT INTO published_road_network (id, created) VALUES (PUBLISHED_ROAD_NETWORK_SEQ.NEXTVAL, sysdate)""".execute
+  def createPublishedRoadNetwork (networkVersion: Long): Unit = {
```
pedrosag: Small suggestion:

If we are to use Sequences.nextPublishedRoadNetworkId on every call of this, then why not define this as the default value?

PR #456, opened 2019-02-11
Viite 1717 removal of unneeded code about floating and unaddressed links

pedrosag:
Removed all the code I could find about the floatings and the unaddressed.

Current tests are green on jenkins, build 353.


pedrosag:
Do not mind the miss naming of the commits, sometimes InteliJ Îdea has aneurysms....

PR #457, opened 2019-02-11
Viite 1800 direction arrow is not being the same as the one previously existing

Campanha:


Campanha:
http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/355/console

PR #460, opened 2019-02-12
VIITE-1806 Form inputs disabled if project status is not incomplete.

skosonen:


viite-UI/src/view/FormCommon.js
```diff
@@ -26,25 +26,26 @@
...
-    var newRoadAddressInfo = function(selected, links, road){
+    var newRoadAddressInfo = function(project, selected, links, road) {
       var roadNumber = road.roadNumber;
       var part = road.roadPartNumber;
       var track = road.trackCode;
       var roadName = selected[0].roadName;
       var link = _.first(_.filter(links, function (l) {
         return !_.isUndefined(l.status);
       }));
+      var projectIncomplete = project.statusCode === 1;
```
hestrompa: We should avoid this magic number

skosonen: True, I'll fix it!

PR #467, opened 2019-02-15
Viite 1788 in the end of road discontinuity code should be 1

Campanha:


Campanha:
all tests and new passing: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/368/console

PR #470, opened 2019-02-19
Viite 1805 bug in automatic reversion v2

Campanha:


Campanha:
all tests pass: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/378/console

PR #474, opened 2019-02-21
Viite 1803 direction arrows are pointing different direction

hestrompa:


hestrompa:
Tests are ok: [](http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/382/console)

PR #476, opened 2019-02-22
VIITE-1812 minor discontinuities on roundabout

hestrompa:
- increasing MaxDistanceForSearchDiscontinuityOnOppositeTrack limit to 20 meters.
- Measure of distance between track endpoints is done based on geometry distances, instead of addressMValues in meters


Campanha:
all tests pass : http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/385/consoleFull

PR #480, opened 2019-02-26
Viite 1833 road checker exception fix

elenaols:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadNetworkService.scala
```diff
@@ -89,8 +89,7 @@ class RoadNetworkService {
         } else {
           val firstOpt = allLocations.find(loc => loc.orderNumber == 1 && loc.calibrationPoints._1.nonEmpty && loc.calibrationPoints._1.get == 0)
           val lastOpt = Some(allLocations.maxBy(_.calibrationPoints._2))
-
-          if (firstOpt.isEmpty && lastOpt.isEmpty) {
+          if (firstOpt.isEmpty || lastOpt.get.endCalibrationPoint.isEmpty || !lastOpt.get.endCalibrationPoint.contains(Some(roadways.maxBy(_.endAddrMValue)).get.endAddrMValue)) {
```
skosonen: What if lastOpt is None? This would throw that "None.get" exception then.

elenaols: In firstOpt we use 'find' and there is a possibility to get None, but in lastOpt we use Some Option that returns lastOpt options also in cases, when there are no linear locations with end m values (they just had none in endCalibration point). It returned lastOpt.isEmpty = false also in cases, where there were no end m values (I was wondering it as well). That's why I had to add this 'lastOpt.get.endCalibrationPoint.isEmpty' to search for empty endpoints in lastOption. I tested with cases, when there were no end points and it all looks working OK.

skosonen: So this is fine then.

PR #502, opened 2019-03-07
Viite 1809 remove history addresses if equal or bigger than exported date

Campanha:


Campanha:
tests pass http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/397/console

PR #503, opened 2019-03-07
VIITE-1855 Fixing validation for EndOfRoad in the middle of a part

hestrompa:


Campanha:
tests pass: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/398/console

PR #504, opened 2019-03-08
Viite 1854 roadname can be null if bigger than 70000

Campanha:


Campanha:
tests failing: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/401/console

Campanha:
http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/402/console

PR #509, opened 2019-03-19
Viite 1873 project link name uk violated

Campanha:


Campanha:
tests pass: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/405/console

PR #512, opened 2019-03-20
Viite 1863 give errror when saving roadpart zero

Campanha:


Campanha:
tests passing: http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/421/console

PR #520, opened 2019-04-01
VIITE-1871 minor fix due to firefox and ie incompatibility

andre-policarpo:


skosonen:
Tested locally on Firefox and it worked.

PR #522, opened 2019-04-02
VIITE-1875 Restricting the length of input fields, VIITE-1877 Fix for SQL

skosonen:
Restricting the length of road number to five, road part number to three and ELY to two characters in input fields.


skosonen:
Committed accidentally VIITE-1877 fix in this branch as well.

PR #524, opened 2019-04-02
VIITE-1880 refresh button added

andre-policarpo:
UI Needs to be verified before merging.


andre-policarpo:
Every js I'm writting is failing on Firefox - before merging let me test this please. - Thanks in advance!

PR #526, opened 2019-04-03
Viite 1387 error message changed for new roads

andre-policarpo:
New message added for viite\package.scala: NewRoadAlreadyExistsMessage


andre-policarpo:
Questions:
1. The message: "Antamasi tienumero ja tieosanumero ovat jo käytössä. Tarkista syöttämäsi tiedot." - already existed. But I didn't found it back then. I don't know why. /blame

viite\package.scala:
val ErrorRenumberingValuesAlreadyInUse = "Antamasi tienumero ja tieosanumero ovat jo käytössä. Tarkista syöttämäsi tiedot."

2. I've created a new variable in viite\package.scala:
val NewRoadAlreadyExistsMessage = "Antamasi tienumero ja tieosanumero ovat jo käytössä. Tarkista syöttämäsi tiedot."

Since it's the same message should we rename the old variable to suit both cases or we can stay with both variables? - In my opinion this might be a problem when we want to find out why an error is occurring.

NOTE: If you're OK with both variables, remove "DO NOT MERGE" label and "#push-it baby"

elenaols:
It is better to have only 1 variable

andre-policarpo:
Removed the duplicate message.
Rename the old message to suit both cases.

Fixed tests to assert to the new message.

Tests OK at http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/452/consoleFull

PR #538, opened 2019-04-18
VIITE-1917 Transfer after terminated should be adjusted also in Sources

Campanha:


Campanha:
all tests passing at http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/484/console

PR #540, opened 2019-04-23
Viite 1885 hide numerointi from action drop down

andre-policarpo:
Running tests..


andre-policarpo:
http://172.17.201.56:8888/job/Run%20OTH%20+%20Viite%20Tests/485/consoleFull
[success] Total time: 881 s, completed Apr 23, 2019 1:10:09 PM

PR #549, opened 2019-04-30
Viite 1932 link source in initial import

skosonen:


skosonen:
Tests pass on Jenkins: #494

PR #550, opened 2019-04-30
Viite 1879 deletion of reserved and formed parts

andre-policarpo:
Finished: SUCCESS at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/2/consoleFull


skosonen:
I think this is done according to the description in the ticket, but there is comment from Janne, which changes the whole thing: "**let's do now the simplest fix: removing the bin away formed road parts.**"

andre-policarpo:
> I think this is done according to the description in the ticket, but there is comment from Janne, which changes the whole thing: "**let's do now the simplest fix: removing the bin away formed road parts.**"

I've spoken with Janne back then, that removing the bin away wouldn't make the solution easier - because removing one reserved part would also modify the formed list.

I think I've spent more time than it was suppose for this task, but I also think that now every think is working as intended.

So now, when removing a part, either reserved or formed, it will remove it's lower dependencies - not the upper ones. Which means:
1. removing one reserved part will remove it's formed part, if any.
2. removing one formed part will not remove the reserved entry.

I've also added a flag to mark if a certain part is Renumbered or not - so:
1. when removing a reserved part - it will check in the formed list if there is any formed part that was renumbered from this reserved part we're removing.

One more thing:
I've added a list of roads (roadNumber and roadPartNumber) to the formed list - so now, it is possible to look into the formed part object and see all the roads that are beeing used for this road, like:
1. transfer part 1 and part 2 to part 3:
2. formed part 3 will have an array of roadAddresses = [1, 2], saying that the parts 1 and 2 are now beeing transfered to 3.

I've followed the two cases in the excel file :)

TL;DR : I didn't removed the trash bin, because I've found a working solution :)

PR #557, opened 2019-05-13
VIITE-1893 added fix for first check

Campanha:
Added fix for error message in Transfer cases where the ne…w road part was not reserved in the project


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -83,8 +83,17 @@ class ProjectValidator {
     }
   }
 
-  def checkNotReserved(number: Long, part: Long, currentProject: Project): Unit = {
-    val project = projectReservedPartDAO.roadPartReservedByProject(number, part, currentProject.id, withoutProjectId = true)
+  def checkReservedPartInProject(number: Long, part: Long, currentProject: Project, linkStatus: LinkStatus): Unit = {
```
elenaols: Is it possible to pass LinkStatus as optional parameter and use just one checkReservedPartInOtherProject? And combine ProjectValidationExceptions like:
 if (project.isEmpty) { throw new ProjectValidationException(RoadPartNotReservedInProjectMessage.format(number, part, currentProject.name))
} else {
throw new ProjectValidationException(RoadReservedOtherProjectMessage.format(number, part, currentProject.name)) 
}

Campanha: well ive refactored the DAOs fetch methods a bit but the val project are not the opposite in the functions and the if else will not work n this case

Campanha:
Do not merge: still need to change tests due to new check validations

Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/9/console

PR #564, opened 2019-05-17
Viite 1893 checks when saving action

Campanha:
New Transfer check added


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/13/console

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -65,15 +65,18 @@ class ProjectValidator {
       if (roadPartLinks.exists(rpl => rpl.status == Numbering)) {
         throw new ProjectValidationException(ErrorTransferActionWithNumbering)
       }
+    } else if(LinkStatus.Numbering.value == linkStatus.value){
```
andre-policarpo: } else if (LinkStatus.Numbering.value == linkStatus.value) { // spacing

PR #567, opened 2019-05-20
VIITE-1958 links with Ennallaan status should not be after any other …

Campanha:
…link with different status (other than NotHandled)


Campanha:
missing new unit test

Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/19/console

PR #573, opened 2019-05-23
VIITE-1975 numerointi hotfix

andre-policarpo:
don't deploy this fix before testin'.

Everything seems to work now.

### DIFF
--- a/viite-UI/src/model/ProjectCollection.js
+++ b/viite-UI/src/model/ProjectCollection.js
@@ -117,6 +117,8 @@
           publishable: result.publishable
         };
         projectErrors = result.projectErrors;
+        reservedParts = result.reservedInfo;
+        formedParts = result.formedInfo;
         publishableProject = result.publishable;
         eventbus.trigger('roadAddressProject:projectFetched', projectInfo);
       });
@@ -196,10 +198,11 @@
             startDate: result.project.startDate,
             publishable: false
           };
-          projectErrors = result.projectErrors;
-          eventbus.trigger('roadAddress:projectSaved', result);
-          reservedPartList = result.reservedInfo;
           currentProject = result;
+          projectErrors = result.projectErrors;
+          reservedParts = result.reservedInfo;
+          formedParts = result.formedInfo;
+          eventbus.trigger('roadAddress:projectSaved', result);
         }
         else {
           eventbus.trigger('roadAddress:projectValidationFailed', result.errorMessage);
@@ -279,6 +282,7 @@
                 } else {
                   publishableProject = successObject.publishable;
                   projectErrors = successObject.projectErrors;
+                  formedParts = successObject.formedInfo;
                   eventbus.trigger('projectLink:projectLinksCreateSuccess');
                   eventbus.trigger('roadAddress:projectLinksUpdated', successObject);
                 }
@@ -288,6 +292,7 @@
                 if (successObject.success) {
                   publishableProject = successObject.publishable;
                   projectErrors = successObject.projectErrors;
+                  formedParts = successObject.formedInfo;
                   eventbus.trigger('roadAddress:projectLinksUpdated', successObject);
                 } else {
                   new ModalConfirm(successObject.errorMessage);
@@ -671,12 +676,6 @@
       });
       _.each(qRoadParts, function (row) {
         reservedParts.push(row);
-      });
-    };
-
-    this.deleteRoadPartFromList = function (list, roadNumber, roadPartNumber) {
-      return _.filter(list, function (dirty) {
-        return !(dirty.roadNumber.toString() === roadNumber && dirty.roadPartNumber.toString() === roadPartNumber);
       });
     };
 
--- a/viite-UI/src/model/ProjectCollection.js
+++ b/viite-UI/src/model/ProjectCollection.js
@@ -229,6 +229,7 @@
             dirtyProjectLinkIds = [];
             publishableProject = response.publishable;
             projectErrors = response.projectErrors;
+            formedParts = response.formedInfo;
             eventbus.trigger('projectLink:revertedChanges');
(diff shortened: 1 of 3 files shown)


andre-policarpo:
For this _Numerointi_ option to be hidden when is not possible: I've matched the road number and road part number with the ones already modified (formed list):
* If the road part and road part number is present in the modified roads (formed list), then the option _Numerointi_ is not possible;
Here are an explanation of the bugs found and what was fixed:
1. It was detected that after transferring a link the _Numerointi_ option is still appearing on the dropdown:
* **Reason**: The reason this was happening it's because the _formed list_ wasn't being updated.
* **What was done**: I added an update after the user does any change in project links - so the _formed list_ keeps synchronized with the data base.
* **Current situation**: This update fixed both cases of transferring and reverting this same transfer, at this moment the _Numerointi_ option is properly being hidden when it should and otherwise.
2. It was detected that after revert the renumbering a road the _Numerointi_ option is still appearing on the dropdown for a while:
* **Reason**: As you know, the spin is added whenever the data is being loaded, the problem here is that the spin was beeing removed sooner than expectded.
* **What was done**: Delay the spin remove.
* **Current situation**: I hope this changes are good enough - I still need this to be tested in devtest since the environment is faster and we can see some different behaviors happening.

viite-UI/src/model/ProjectCollection.js
```diff
@@ -117,6 +116,8 @@
           publishable: result.publishable
         };
         projectErrors = result.projectErrors;
```
Campanha: we can start to use the existing methods setProjectErrors, setReservedParts and setFormedParts

viite-UI/src/model/ProjectCollection.js
```diff
@@ -196,10 +197,11 @@
             startDate: result.project.startDate,
             publishable: false
           };
+          currentProject = result;
           projectErrors = result.projectErrors;
+          reservedParts = result.reservedInfo;
```
Campanha: we can start to use the existing methods setProjectErrors, setReservedParts and setFormedParts

viite-UI/src/model/ProjectCollection.js
```diff
@@ -269,34 +271,37 @@
...
+          resetEditedDistance();
+          var ids = dataJson.ids;
+          if (dataJson.linkStatus === LinkStatus.New.value && ids.length === 0) {
+            backend.createProjectLinks(dataJson, function (successObject) {
+              if (!successObject.success) {
+                new ModalConfirm(successObject.errorMessage);
+                applicationModel.removeSpinner();
+              } else {
+                publishableProject = successObject.publishable;
+                projectErrors = successObject.projectErrors;
```
Campanha: same as previous: we can start to use the existing methods setProjectErrors, setReservedParts and setFormedParts

andre-policarpo: For this _Numerointi_ option to be hidden when is not possible: I've matched the road number and road part number with the ones already modified (formed list):
 * If the road part and road part number is present in the modified roads (formed list), then the option _Numerointi_ is not possible;

Here are an explanation of the bugs found and what was fixed:
 1. When transferring a link, the _Numerointi_ option is still appearing on the dropdown:
 * **Reason**: The reason this was happening it's because the _formed list_ wasn't being updated.
 * **What was done**: I added an update after the user does any change in project links - so the _formed list_ keeps synchronized with the database.
 * **Current situation**: This update fixed the synchronization of the _formed list_ when any change is made. Now the _Numerointi_ option is being hidden when it should.

2. When reverting the renumbering, the _Numerointi_ option is still hidden:
 * **Reason*: As you know, the spin is added whenever the data is being loaded, the problem here is that the spin was beeing removed sooner than expectded.
 * **What was done**: Delayed when the spin is removed.
 * **Current situation**: I hope this changes are good enough - I still need this to be tested in devtest since the environment is faster and we can see some different behaviors happening.

3. When reverting any other change, the Numerointi option is still hidden:
 * *Reason*: The reason this was happening it's because the _formed list_ wasn't being updated.
 * *What was done*: I added an update after the user does any revert change in project links - so the _formed list_ keeps synchronized with the data base.
 * **Current situation** (similar with point **1.**): This update fixed the synchronization of the _formed list_ when any change is reverted. Now the _Numerointi_ option is visible as it should.

andre-policarpo:
For this _Numerointi_ option to be hidden when is not possible: I've matched the road number and road part number with the ones already modified (formed list):
 * If the road part and road part number is present in the modified roads (formed list), then the option _Numerointi_ is not possible;

Here are an explanation of the bugs found and what was fixed:
 1. When transferring a link, the _Numerointi_ option is still appearing on the dropdown:
 * **Reason**: The reason this was happening it's because the _formed list_ wasn't being updated.
 * **What was done**: I added an update after the user does any change in project links - so the _formed list_ keeps synchronized with the database.
 * **Current situation**: This update fixed the synchronization of the _formed list_ when any change is made. Now the _Numerointi_ option is being hidden when it should.

2. When reverting the renumbering, the _Numerointi_ option is still hidden:
 * **Reason*: As you know, the spin is added whenever the data is being loaded, the problem here is that the spin was beeing removed sooner than expectded.
 * **What was done**: Delayed when the spin is removed.
 * **Current situation**: I hope this changes are good enough - I still need this to be tested in devtest since the environment is faster and we can see some different behaviors happening.

3. When reverting any other change, the Numerointi option is still hidden:
 * *Reason*: The reason this was happening it's because the _formed list_ wasn't being updated.
 * *What was done*: I added an update after the user does any revert change in project links - so the _formed list_ keeps synchronized with the data base.
 * **Current situation** (similar with point **1.**): This update fixed the synchronization of the _formed list_ when any change is reverted. Now the _Numerointi_ option is visible as it should.

PR #577, opened 2019-05-28
VIITE-1967 added order by municipaliy and fixed distance NaN

andre-policarpo:


andre-policarpo:
During this task, I've found a bug regarding the objects in the array:
There is this field, 'distance' that was always _NaN_.
Before this list was being sorted by this 'distance' field, since it was _NaN_ - the list was kept the same.

This bug was added in _2017_, and was never noticed.
**FYI**: '_event.center_' doesn't returns the same as '_event.bbox.getCenterLonLat()_'

**P.S.**: I've kept this 'distance' anyway, but I think that if we can't explain it we should remove it.

PR #590, opened 2019-06-05
VIITE-1984 fixing update for numbering operation

hestrompa:


viite-UI/src/view/ProjectEditForm.js
```diff
@@ -282,9 +282,11 @@
         formCommon.replaceAddressInfo(backend, selectedProjectLink);
         checkInputs('.project-');
```
skosonen: Duplicate code could be refactored to a common function.

viite-UI/src/view/ProjectEditForm.js
```diff
@@ -256,9 +254,11 @@
         formCommon.replaceAddressInfo(backend, selectedProjectLink, currentProject.project.id);
         checkInputs('.project-');
```
skosonen: Duplicate code could be refactored to a common function.

PR #594, opened 2019-06-12
VIITE-1899 Validation points to the first project link with different…

skosonen:
… road type instead of first project link.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -577,27 +577,25 @@ class ProjectValidator {
...
       }
-
-      val rightRoadTypes = right.sortBy(_.startAddrMValue).map(_.roadType.value).foldLeft(Seq.empty[Int]) { case (list, next) =>
-        if (list.nonEmpty && list.last == next)
-          list
-        else
-          list :+ next
+      if (!leftIter.hasNext && !rightIter.hasNext) {
+        None
+      } else if (sortedLeft.size > sortedRight.size) {
```
Campanha: do we want to check sizes of the list? because in two tracks its normal to have more links in one side than in the other, which will create false positive error messages for this cases

PR #600, opened 2019-06-27
Viite 1981 junction hover

elenaols:
- Junction address to one row
- Node name added


elenaols:
I left Service and Dao getNodeById in case if we will need them in future

PR #601, opened 2019-06-27
Viite 1954 Library upgrades

hestrompa:
This PR includes all library upgrades for backend and frontend


src/main/scala/fi/liikennevirasto/digiroad2/DigiroadServer.scala
```diff
@@ -73,11 +72,11 @@ class OAGRasterServiceProxyServlet extends ProxyServlet {
   private val logger = LoggerFactory.getLogger(getClass)
   def regex: Regex = "/(viite)".r
 
-  override def rewriteURI(req: HttpServletRequest): java.net.URI = {
+  override def rewriteTarget(req: HttpServletRequest): String = {
     val uri = req.getRequestURI
     val url = "http://oag.liikennevirasto.fi" + regex.replaceFirstIn(uri, "") + "?" + req.getQueryString
```
Campanha: we should probably now use vayla instead of liikennevirasto

src/main/scala/fi/liikennevirasto/digiroad2/DigiroadServer.scala
```diff
@@ -57,10 +57,9 @@ class OAGProxyServlet extends ProxyServlet {
 
   def regex: Regex = "/(viite)".r
 
-  override def rewriteURI(req: HttpServletRequest): java.net.URI = {
+  override def rewriteTarget(req: HttpServletRequest): String = {
     val uri = req.getRequestURI
-    java.net.URI.create("http://oag.liikennevirasto.fi/rasteripalvelu-mml"
-      + regex.replaceFirstIn(uri, ""))
+    "http://oag.liikennevirasto.fi/rasteripalvelu-mml" + regex.replaceFirstIn(uri, "")
```
Campanha: we should probably now use vayla instead of liikennevirasto

Campanha:
> under construction links are not being properly filtered

this is now fixed in new branch with current devtest code

PR #605, opened 2019-07-04
VIITE-2011 fix calibration points recalculation

hestrompa:
- When roadway_number changes, we should not add a calibration point


hestrompa:
Tests are green: [Tests](http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/27/console)

PR #612, opened 2019-07-15
VIITE-1948 KMTK Database changes.

skosonen:


skosonen:
Please notice, that the test data has not been changed and initializing db with the test data will fail.

PR #621, opened 2019-07-30
VIITE-1899 select links that has different road types

andre-policarpo:
Now _Korja_ button selects all the links that have different road type for the same addresses in both tracks. Even if one of the tracks is still unchanged.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
@@ -577,35 +577,17 @@ class ProjectValidator {
     }
 
     def checkMinMaxTrackRoadTypes(trackInterval: Seq[ProjectLink]): Option[Seq[ProjectLink]] = {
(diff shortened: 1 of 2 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -576,28 +576,35 @@ class ProjectValidator {
...
-          list :+ next
+          list :+ (next.roadType.value, next.startAddrMValue, next.endAddrMValue)
       }
 
-      if (!(leftRoadTypes sameElements rightRoadTypes)) {
-        if (left.nonEmpty)
-          Some(left.head)
-        else
-          Some(right.head)
+      if (!(leftRoadTypes equals rightRoadTypes)) {
```
kohonenm: Now we are checking links two times in case there are different amount of links on different tracks. Would it be enough to use this second algorithm to check all cases: roads with equal amount of links on both tracks and roads with different amount of links on different tracks?

andre-policarpo: Done.

PR #622, opened 2019-07-31
VIITE-1995 info text added

andre-policarpo:
Added text information to the user for when the search is running.


kohonenm:
Jenkins test ok, http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/72/

PR #623, opened 2019-08-02
Viite 2035 reserved road part is kept when all links are transferred

andre-policarpo:
_newSavedLinks_ are now beeing filtered to include only newLinks. To prevent an existent road part to be removed from the reserved parts in project.


kohonenm:
Test in Jenkins ok, http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/71/

PR #627, opened 2019-08-06
VIITE-2036 added node tool buttons in the left panel

andre-policarpo:


andre-policarpo:
**What was done:**

1. Now buttons do not appear while in _roadAddressProject_ layer or _linkProperty_ layer
2. Buttons are shown when in _node_ layer
3. When the highlighted button is clicked, the selection is removed
4. Can only be highlighted one button at a time, either _Select_ or _Add_

PR #628, opened 2019-08-08
Viite 2024 problem with roadway number

Campanha:
RoadwayNumbers for New project links should be generated together, if applicable (in the same continuous section)


Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/70/consoleFull

PR #638, opened 2019-08-23
VIITE-1949 KMTK initial import

skosonen:


skosonen:
Tests will not pass, but initial import seemed to work well from KMTK_TIEOSOITE_JOULUKUU_2018 table.

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -231,18 +232,19 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
...
-      projectService.fetchPreFillFromVVH(linkId, currentProjectId) match {
+      projectService.fetchPreFill(linkId, currentProjectId) match {
         case Right(preFillInfo) =>
           Map("success" -> true, "roadNumber" -> preFillInfo.RoadNumber, "roadPartNumber" -> preFillInfo.RoadPart, "roadName" -> preFillInfo.roadName, "roadNameSource" -> preFillInfo.roadNameSource.value)
         case Left(failureMessage) => Map("success" -> false, "reason" -> failureMessage)
       }
     }
   }
 
-  get("/roadlinks/midpoint/:linkId") {
```
skosonen: Maybe we could keep using linkIds in the UI even though those are generated by us? Or do we want to let the user put uuid / version in the URL the same way the user can now put linkId?

skosonen: Got answer: we need to show the UUID/Version.

PR #640, opened 2019-08-29
VIITE-2031 Fixing calibration points update

hestrompa:


Campanha:
all tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/84/console

PR #643, opened 2019-09-03
VIITE-2061 Dif. Road types mixes the change table on new 2-track roads

Campanha:
Adjust between both tracks should be made according to their "rowchanges" ascending order.


Campanha:
all tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/83/consoleFull

PR #644, opened 2019-09-04
Viite 2034 Form of Junction

kohonenm:


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -1166,6 +1180,15 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     )
   }
 
+  def junctionInfoToApi(junctionInfo: JunctionInfo) : Map[String, Any] = {
+      Map("id" -> junctionInfo.id,
+        "junctionNumber" -> junctionInfo.junctionNumber,
+        "nodeId" -> junctionInfo.nodeId,
+        "startDate" -> formatDateTimeToShortPatternString(Some(junctionInfo.startDate)),
```
andre-policarpo: Just a quick note: **formatDateTimeToShortPatternString** was already define as **formatToString**.

Although I think that this implementation is better since it receives an actual DateTime.
I think we can try to use only this function on other places, to avoid two functions that do the same thing.

Then, formatToString can be removed.

viite-UI/src/less/core/forms.less
```diff
@@ -287,10 +287,19 @@ form.roadAddressProject .control-label-small {
   margin-right: 11px;
   margin-top: 30px;
   margin-bottom: -10px;
-  width: 55px !important;
+  width: 45px !important;
```
andre-policarpo: I'm afraid that modifying classes already in use by other forms : just to adapt to a newly created form is not a good idea.
We should have one or two classes : junction / junction point and there we define all the rules we need.

Even if the rules are the same, it will be alot easier to modify them only for one specific form in the future.

viite-UI/src/view/JunctionEditForm.js
```diff
@@ -0,0 +1,154 @@
+(function (root) {
+    root.JunctionEditForm = function (backend) {
+        var svgJunction =
```
andre-policarpo: svg pictures are already defined. I think you can use them like this : 

`> <object type="image/svg+xml" data="images/**filename**.svg>`

viite-UI/src/view/NodeLayer.js
```diff
@@ -69,6 +69,141 @@
...
+      junctionPointTemplateClick.on('select', function (event) {
+        // var selected = _.find(event.selected, function (selectionTarget) {
+        //   return !_.isUndefined(selectionTarget.nodePointTemplateInfo);
+        // });
+        // sets selected mode by default - in case a node point is clicked without any mode
+        var selected = _.find(event.selected, function (selectionTarget) {
+          return !_.isUndefined(selectionTarget.junctionPointTemplateInfo);
+        });
+        eventbus.trigger('junctionEdit:selected', selected.junctionPointTemplateInfo.junctionId);
+        // if (!_.isUndefined(selected) && applicationModel.selectedToolIs(LinkValues.Tool.Unknown.value)) {
```
andre-policarpo: Check this comments/file to remove unnecessary code/comments.

viite-UI/src/less/core/forms.less
kohonenm: Fixed and pushed to Git.

viite-UI/src/view/JunctionEditForm.js
kohonenm: Fixed and pushed to Git.

viite-UI/src/view/NodeLayer.js
kohonenm: Fixed and pushed to Git.

PR #645, opened 2019-09-04
Viite 2039 form of junction points

kohonenm:


viite-UI/src/less/core/forms.less
```diff
@@ -290,7 +290,17 @@ form.roadAddressProject .control-label-small {
   width: 55px !important;
   text-align: center;
 }
+form.roadAddressProject .junction-control-label-small{
```
andre-policarpo: This form is to display junction points - shouldn't be linked to the roadAddressProject class.

viite-UI/src/view/JunctionEditForm.js
```diff
@@ -0,0 +1,156 @@
+(function (root) {
+    root.JunctionEditForm = function (backend) {
+        var junctionIdForLink = 0;
+        var svgJunction =
```
andre-policarpo: svg pictures are already defined. I think you can use them like this : 

`> <object type="image/svg+xml" data="images/**filename**.svg>`

I'm duplicating comments - we need to be carefull with merging this user-stories since they have a lot of duplicated code.

kohonenm: Fixed in Viite-2034 branch, where JunctionEditForm.js is implemented

viite-UI/src/less/core/forms.less
kohonenm: Fixed with new commit

PR #658, opened 2019-09-18
VIITE-2062 Not re-calculating addresses for terminated parts.

skosonen:


Campanha:
is it possible that we could have same problem in ProjectSectionMValuesCalculator line 80 ?
Is same method name different signature

skosonen:
@Campanha I think that is used for a different purpose and is called before the function I changed. I think that no need to change that.

Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/89/console

PR #659, opened 2019-09-19
Viite 2049 addr values after road type change

Campanha:
Problem1 : Calculator strategies where prepared since a long time not to change Unchanged links i. e. the code was not prepared to update addressing M values of links that were not supposed to be updated.Code was changed due to RoadType djustment not in all places as it seems.

Problem 2: Also before entering this calculator proccess, the links were having wrong order since they were being sorted by starting Wrong addresses.
Now they are being sorted by geometry.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionMValueCalculator.scala
```diff
@@ -51,8 +51,7 @@ object ProjectSectionMValueCalculator {
       val addressValue = if (someCalibrationPoint.nonEmpty) someCalibrationPoint.get.addressMValue else m + pl.geometryLength * coEff
       pl.status match {
         case LinkStatus.New => addressValue
-        case LinkStatus.Transfer | LinkStatus.NotHandled | LinkStatus.Terminated | LinkStatus.Numbering => m + pl.addrMLength
-        case LinkStatus.UnChanged => pl.endAddrMValue
+        case LinkStatus.Transfer | LinkStatus.NotHandled | LinkStatus.Terminated | LinkStatus.Numbering | LinkStatus.UnChanged => m + pl.addrMLength
```
skosonen: VIITE-2062 has also change for this place. Terminated would be like UnChanged was here before.

Campanha: yep i noticed that also

Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/90/console

PR #678, opened 2019-10-08
Viite 2066 and Viite 2084 related with node information

andre-policarpo:


viite-UI/src/less/core/forms.less
```diff
@@ -373,6 +362,40 @@ form.roadAddressProject p {
...
+  color: #fff;
+  font-weight: bold;
+  font-size: 12px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  text-align: left;
+}
+
+.node-junctions-table-border-bottom {
+  border-bottom: 1px dashed @gray-light;
```
skosonen: I think these separating lines would be easier for eyes if they were solid instead of dashed.

PR #679, opened 2019-10-09
Removal of static capistrano version in deploy.

Campanha:


skosonen:
What about the Gemfile:
`gem 'capistrano', '~> 3.11.1'`

PR #685, opened 2019-10-15
VIITE-2118 removed wrong list iteration concatenation

Campanha:


Campanha:
all tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/94/console

PR #688, opened 2019-10-15
VIITE-2126 keep node selected after drag and when selecting another node

andre-policarpo:


Campanha:
before merge please delete logger.info(s"UpdateRodways log 1") and logger.info(s"UpdateRodways log 2") from ProjectService.scala line 1866. I forgot to delete those.

PR #691, opened 2019-10-16
Viite 1954 library updates with devtest integration

Campanha:


Campanha:
Fixture reset OK: http://livibuild04.vally.local/jenkins/job/viite/job/Fixture-Reset/145/console
Tests OK: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/95/console

PR #706, opened 2019-11-06
VIITE-2060 Showing unique list of the node point templates.

skosonen:
Because of dual points there can be two node points in the same address, but in the list we just show one link.

### DIFF
--- a/viite-UI/src/view/NodeSearchForm.js
+++ b/viite-UI/src/view/NodeSearchForm.js
@@ -112,7 +112,10 @@
     };
 
     var nodePointTemplatesHtml = function (nodePointTemplates) {
-      var groups = _.groupBy(nodePointTemplates, function (template) {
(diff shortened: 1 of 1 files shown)


viite-UI/src/view/NodeSearchForm.js
```diff
@@ -112,7 +112,10 @@
     };
 
     var nodePointTemplatesHtml = function (nodePointTemplates) {
-      var groups = _.groupBy(nodePointTemplates, function (template) {
+      var uniqueNPTemplates = _.uniqWith(nodePointTemplates, function (o1, o2) {
```
andre-policarpo: I would write this as something like:
`_.uniqBy(nodePointTemplates, ['roadNumber', 'roadPartNumber', 'addrM'])`

skosonen: This gives only one item in list so, reverted and using original uniqWith instead.

PR #707, opened 2019-11-06
Viite 2160 give new roadway numbers when addr change

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/113/console

PR #709, opened 2019-11-11
Viite 2146 two track road is missing row(s) in change tbl

kohonenm:


kohonenm:
Changes according to Viite needs to have discontinuity code 6 when link is a parallel link to discontinuous link in two-track roads. Only at the end when sending changes to TR this discontinuity code 6 is mapped to 5.

viite-UI/src/view/ProjectChangeTable.js
```diff
@@ -276,7 +276,12 @@
       target.setAttribute('data-y', y);
     }
 
-
+    function replaceParallelLink(currentDiscontinuity) {
+      if (currentDiscontinuity === 6)
```
Campanha: we should create Discontinuity is LinkValues.js and use its values

PR #710, opened 2019-11-12
Viite 2139 remove 'tieosoiteprojektit'  btn ,'aloita valitsemalla projekti' txt

kohonenm:


kohonenm:
Wrong notification text in commit 0e97407 "VIITE-2146 two track road is missing row(s) in Change table." Ignore that.

viite-UI/src/view/LinkPropertyForm.js
```diff
@@ -450,14 +450,17 @@
...
           '<button id="formNodesAndJunctionsButton" class="open-tool-mode-btn btn btn-block btn-primary" style="margin-top: 5px;">Solmut ja liittymät</button>' +
           '</div>';
         rootElement.append(emptyFormDiv);
         $('[id=formProjectButton]').click(function () {
-          $('[id=projectListButton]').click();
+          if (applicationModel.isProjectOpen()) {
+            new ModalConfirm("Projektin muokkaus on kesken. Tallenna muutokset ja/tai poistu Peruuta-painikkeella.");
+          } else {
+            projectListModel.show();
+          }
```
andre-policarpo: I believe we could change this to:
```
projectListModel.show();
```

Since it won't be possible anymore to see this button with a project open.
This code was used for the other button, because we could open another project with another project still open.

Or, we can leave it like this just for safety, although in normal situation `applicationModel.isProjectOpen()` will never return **true**.

kohonenm:
Requested changes fixed, LinkPropertyForm.js stay as it is for safety reasons. Could be changed, if needed.

viite-UI/src/router.js
```diff
@@ -71,7 +71,7 @@
         eventbus.trigger('underConstructionRoads:toggleVisibility', false);
         $('.underconstruction-visible-wrapper').hide();
         $('#toggleEditMode').hide();
-        $('#emptyFormDiv,#projectListButton').hide();
+        $('#emptyFormDiv,#formProjectButton').hide();
```
andre-policarpo: Can stay as: `$('#emptyFormDiv').hide();`

I think _historyLayer_ is not being used at this moment.

If so, I think we should ask Janne if we want to hide the main project button (on the right)

viite-UI/src/view/NodeLayer.js
```diff
@@ -539,12 +539,12 @@
     var showLayer = function () {
       me.start();
       me.layerStarted(me.eventListener);
-      $('#projectListButton').prop('disabled', true);
+      $('#formProjectButton').prop('disabled', true);
```
andre-policarpo: Can be removed. The button is not visible anyway.

viite-UI/src/view/NodeLayer.js
```diff
@@ -539,12 +539,12 @@
...
     };
 
     var hideLayer = function () {
       me.clearLayers(layers);
-      $('#projectListButton').prop('disabled', false);
+      $('#formProjectButton').prop('disabled', false);
```
andre-policarpo: Can be removed. The button is not visible anyway.

PR #721, opened 2019-11-20
Viite 2191 impossible to create new node on top of templates

andre-policarpo:


andre-policarpo:
Node layer had a bug.
I was working on setting up and down onClick functionalities, and somehow, it wasn't toggle on from start. So user couldn't open node form.

Now it works fine.

PR #737, opened 2019-11-27
VIITE-2204 removed filter that ignore road addresses that are under c…

Campanha:
…onstruction

### DIFF
--- a/viite-UI/src/model/RoadCollection.js
+++ b/viite-UI/src/model/RoadCollection.js
@@ -171,7 +171,7 @@
           });
 
         var nonUnderConstructionRoadLinkGroups = _.reject(roadLinkGroups, function(group) {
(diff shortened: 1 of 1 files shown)


viite-UI/src/model/RoadCollection.js
```diff
@@ -171,7 +171,7 @@
           });
 
         var nonUnderConstructionRoadLinkGroups = _.reject(roadLinkGroups, function(group) {
```
skosonen: Should this variable name be changed since this is not "nonUnderConstructionRoadLinkGroups" any more?

Campanha: true

skosonen: Also good to make sure that on line 177 we don't concatenate same links multiple times.

Campanha: yes we dont get same links since they are being filtered beforehand

PR #738, opened 2019-11-27
VIITE-2208 Prevent selection of unreserved addressed under construction links

skosonen:
In project it is now possible to select unaddressed under construction links but not addressed


skosonen:
Removed check for floating too, since we don't have floating any more.

PR #741, opened 2019-12-03
VIITE-2211 Using VVH timestamp in initial import.

skosonen:


skosonen:
Tests pass on Jenkins. (124)

PR #744, opened 2019-12-04
Viite 2212 nodes and junctions list doesn't zoom right place in every case

kohonenm:


Campanha:
the commit d408cc2 from VIITE-2214 is still under development and it makes one test fail so it should probably be removed otherwise it will fail in tests

PR #747, opened 2019-12-05
VIITE-2231. Viite gives wrong error message if Tierekisteri trrest ap…

kohonenm:
…plication is down

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -1553,7 +1553,11 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
             setProjectStatus(projectId, Sent2TR, withSession = false)
             logger.info(s"Sending to TR successful: ${trProjectStateMessage.reason}")
             PublishResult(validationSuccess = true, sendSuccess = true, Some(trProjectStateMessage.reason))
-
+          case 500 =>
+            logger.error(s"Sending project ${trProjectStateMessage.projectId} to TR failed. Status: ${trProjectStateMessage.status} Error message: ${trProjectStateMessage.reason}")
+            val returningMessage = if (trProjectStateMessage.reason.nonEmpty) trProjectStateMessage.reason else TrConnectionError
+            projectDAO.updateProjectStatus(projectId, SendingToTR)
+            PublishResult(validationSuccess = true, sendSuccess = false, Some(returningMessage))
           case _ =>
             //rollback
             logger.error(s"Sending project ${trProjectStateMessage.projectId} to TR failed. Status: ${trProjectStateMessage.status} Error message: ${trProjectStateMessage.reason}")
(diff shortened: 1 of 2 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ViiteTierekisteriClient.scala
```diff
@@ -221,7 +221,10 @@ object ViiteTierekisteriClient {
     val response = client.execute(request)
     try {
       val statusCode = response.getStatusLine.getStatusCode
-      if (statusCode >= 500) {
+      if (statusCode == 500){
+        val errorMessage = parse(StreamInput(response.getEntity.getContent)).extractOpt[TRErrorResponse].getOrElse(TRErrorResponse("")) // would be nice if we didn't need case class for parsing of one attribute
+        ProjectChangeStatus(trProject.id, statusCode, errorMessage.error_message)
+      } else if (statusCode > 500) {
         logger.info(scala.io.Source.fromInputStream(response.getEntity.getContent).getLines().mkString("\n"))
         throw new RuntimeException("Unable to submit: Tierekisteri error 500")
```
Campanha: should we change message since in this case the error code is not 500?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1553,7 +1553,11 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
             setProjectStatus(projectId, Sent2TR, withSession = false)
             logger.info(s"Sending to TR successful: ${trProjectStateMessage.reason}")
             PublishResult(validationSuccess = true, sendSuccess = true, Some(trProjectStateMessage.reason))
-
+          case 500 =>
+            logger.error(s"Sending project ${trProjectStateMessage.projectId} to TR failed. Status: ${trProjectStateMessage.status} Error message: ${trProjectStateMessage.reason}")
+            val returningMessage = if (trProjectStateMessage.reason.nonEmpty) trProjectStateMessage.reason else TrConnectionError
+            projectDAO.updateProjectStatus(projectId, SendingToTR)
```
Campanha: the main difference seems to be the update in case there is error 500. So in case we got no connection to TR the trProjectStateMessage.reason will be empty. Could we then change the **case _** to 
val returningMessage = if (trProjectStateMessage.reason.nonEmpty) trProjectStateMessage.reason else {
projectDAO.updateProjectStatus(projectId, SendingToTR)
TrConnectionError
} ?

kohonenm: I think there is a difference: there can be connection available to TR, but not available TR execution, then TR returns that 500 in response object. And that response. might have some additional information ?
If there is no connection available ,no response object is returned , then catch() handles that situation ?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ViiteTierekisteriClient.scala
kohonenm: Changed text to " > 500"

PR #750, opened 2019-12-05
Viite 2214 road reverses after adding new link

Campanha:
all tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/126/consoleFull


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/DefaultSectionCalculatorStrategy.scala
```diff
@@ -255,6 +255,88 @@ class DefaultSectionCalculatorStrategy extends RoadAddressSectionCalculatorStrat
...
+//              startPoint1
+//            } else {
+//              chainEndPoints.maxBy(p => direction.dot(p._1.toVector - midPoint))
+//            }
+//          }
+//        }
+//      }
+//    )
+//  }
+
```
andre-policarpo: This old code should be removed.

PR #751, opened 2019-12-05
VIITE-2161. Viite can save too long road names.

kohonenm:


skosonen:
I think Viite database should also have the same 50 characters limit for the road name as in TR. Flyway script could be added in this pull request that would change the max length of the road name.

skosonen:
Devtest, QA and prod databases are ready for shortening the road_name field. (There are no names that are longer than 50 characters.)

PR #752, opened 2019-12-06
VIITE-2214 added validation for two new links before existing road wh…

Campanha:
…ere the closeste new link is not connected but instead it have discontinuity


Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/128/console

PR #760, opened 2019-12-30
VIITE-2149 hotfix to sorting in change table

andre-policarpo:


andre-policarpo:
Changed the way the sort is being done. Now it seems to work Ok.

PR #761, opened 2019-12-31
Viite 2187 detaches  all node points and junction at the same location

andre-policarpo:


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -1253,7 +1253,8 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     Map("id" -> junction._1.id,
       "junctionNumber" -> junction._1.junctionNumber.orNull,
       "nodeNumber" -> junction._1.nodeNumber,
-      "junctionPoints" -> junction._2.map(junctionPointToApi))
+      "junctionPoints" -> junction._2.map(junctionPointToApi)
+    )
```
andre-policarpo: Will fix this.

PR #766, opened 2020-01-08
VIITE-2241 Change table reversed check mark is now bigger

skosonen:
and shown only for the reversed rows. Reversed check mark size increased on the form.


skosonen:
> I'm fine with this, although we should start avoid style rules within _HTML_, and use _CSS_ for this.
> It will make a lot easier to change it in the future, if needed.
> 
> Currently we've a lot of style rules within forms/labels and if we want to change them all, it will be harder/expendable. Like up coming tasks like: VIITE-2152 for example.

Fixed this.

PR #767, opened 2020-01-09
VIITE-2259 Calculated node point checkboxes are now hidden on form.

skosonen:


viite-UI/src/less/core/forms.less
```diff
@@ -716,3 +716,7 @@ form.junction-points .junction-control-label-small{
 form.roadAddressProject .junction-label-projects.header {
   color: @warning;
 }
+
+.form .hidden {
+  display: none;
+}
```
andre-policarpo: This is not needed, since hidden is already an _html/css_ property.

viite-UI/src/view/NodeForm.js
```diff
@@ -273,7 +273,7 @@
         });
         var isDetachable = 'title="' + nodePointType.description + '"'; // added for testing purposes, needs to be confirm if this title is a good idea for production env.
         if (_.isEqual(nodePointType, LinkValues.NodePointType.CalculatedNodePoint)) {
-          isDetachable += ' disabled';
+          isDetachable += ' disabled class="hidden"';
```
andre-policarpo: this could be changed to:
`isDetachable += ' disabled hidden';`

PR #770, opened 2020-01-15
Viite 2249 calibration points for junction points

Campanha:


Campanha:
This branch contains only VIITE-2251.
For VIITE-2250 We will get calibration points for junctions from initial import data.

Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/151/console

PR #771, opened 2020-01-21
VIITE-2258 If we cannot find starting point for the road part by cali…

skosonen:
…bration points or unchanged links, try by unhandled links.


skosonen:
Tests pass on Jenkins: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/155/console

PR #775, opened 2020-01-27
Viite 1862 calculating node points

kohonenm:
Should be merged to master at the same time with sub tasks.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodePointDAO.scala
```diff
@@ -338,9 +341,104 @@ class NodePointDAO extends BaseDAO {
...
   }
+  
+  def expireByNodeNumberAndType(node_number: Long, typeId: Long): Unit = {
+    val query =
+      s"""
+        Update NODE_POINT Set valid_to = sysdate where valid_to IS NULL AND node_number = $node_number
+        AND type = $typeId
+      """
+    Q.updateNA(query).execute
+  }
```
andre-policarpo: Didn't we were suppose to remove all previously calculated node points? Instead of saving them as history?

skosonen: valid_from and valid_to are modification history -fields, not history-fields like start_date and end_date. If valid_to is set, the row is logically deleted. We don't really delete rows, we just expire them by setting valid_to = sysdate.

skosonen:
Jussi created another branch that includes these changes.

PR #777, opened 2020-01-28
VIITE-2255 Two-track road direction

skosonen:


skosonen:
Three tests fail now:

Test orderProjectLinksTopologyByGeometry When choosing the same track when there is 2 connected Then links with only one with same track code should be picked *** FAILED ***

Test defaultSectionCalculatorStrategy.findStartingPoints() When adding one (New) link with inverted geometry before the existing (Transfer) road but where the first link was terminated Then the road should still maintain the previous existing direction *** FAILED ***

Test defaultSectionCalculatorStrategy.findStartingPoints() When adding one (New) link before the existing (Transfer) road but where the first link was terminated Then the road should still maintain the previous existing direction *** FAILED ***

skosonen:
Now four tests are failing. Better to merge changes done by Ricardo first and then work on this one.

skosonen:
Now 7 tests fail.

skosonen:
One broken test left: ProjectSectionCalculatorSpec:881

Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/168/consoleFull
(master branch included)

PR #783, opened 2020-01-31
VIITE-2278. Disable save-buttons from nodes and junctions forms in QA…

kohonenm:
… and Prod

### DIFF
--- a/viite-UI/src/view/JunctionEditForm.js
+++ b/viite-UI/src/view/JunctionEditForm.js
@@ -127,6 +127,12 @@
         });
         return false;
       });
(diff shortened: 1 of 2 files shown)


viite-UI/src/view/JunctionEditForm.js
```diff
@@ -127,6 +127,10 @@
         });
         return false;
       });
+      var saveBtnDisabled = (Environment.name() === 'integration' || Environment.name() === 'production');
+      if (!saveBtnDisabled ) {
```
skosonen: This if should be removed.

PR #787, opened 2020-02-07
VIITE-2281 Problem in adding Left track first instead of Right track

Campanha:
Adding Left track first instead of Right track, should close the form even when not calculated properly, due to missing opposite track


Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/173/console

PR #789, opened 2020-02-11
VIITE-2282 Starting point when multiple discontinuities

skosonen:
Implemented unit tests. The actual problem was fixed in another issue.


skosonen:
Tests pass.

PR #793, opened 2020-02-13
Viite 2055 attaching templates to node

andre-policarpo:


viite-UI/src/view/NodeForm.js
```diff
@@ -395,18 +415,22 @@
...
-      if (!saveBtnDisabled && textIsNonEmpty && nodeTypeIsValid && dateIsNonEmpty && selectedNode.isDirty()) {
-        $('.btn-edit-node-save').prop('disabled', !saveBtnDisabled);
-      }
+    var disableAutoComplete = function () {
+      $('#nodeName').attr('autocomplete', 'false');
+      $('#nodeStartDate').attr('autocomplete', 'false');
+    };
+
+    var formIsInvalid = function () {
+      //  TODO check if this is still necessary for this branch:
```
Campanha: the Tallenna button should be enabled to all environments after its implementation right?

PR #796, opened 2020-02-17
VIITE-2289 findChainEndpoints failing test

Campanha:


digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/process/TrackSectionOrderSpec.scala
```diff
@@ -275,7 +275,7 @@ class TrackSectionOrderSpec extends FunSuite with Matchers {
   }
 
   // TODO Fix this test or the code it tests
```
skosonen: TODO-comment should be removed if this test is fixed.

PR #797, opened 2020-02-17
VIITE-1862 create calculated nodepoints with node and project change

peltsi212:


peltsi212:
Jenkins test is ok.

.gitignore
```diff
@@ -13,3 +13,8 @@ project/project/
 yarn.lock
 package-lock.json
 digiroad2.boot.log
+/build/
+.classpath
+.project
+.settings/
+WebContent/
```
andre-policarpo: Is this suppose to be merged? Just checking.

peltsi212: Yes these changes enable to use eclipse. I use eclipse for git stashes

andre-policarpo: Nothing against Eclipse, it was my first ever IDE, but you should really give it a try to IntelliJ IDE! 😄

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -337,4 +337,27 @@ class NodeDAO extends BaseDAO {
       queryList(query)
     }
   }
+
+  def fetchNodeNumbersByProject(projectId: Long): Seq[Long] = {
```
skosonen: Should we join here also the node -table and check that the node.end_date is null and node.valid_to is null.

skosonen:
Tests pass on Jenkins.

skosonen:
All tests are still passing.

PR #798, opened 2020-02-18
VIITE-2055 disabled save button for QA and Production

andre-policarpo:


viite-UI/src/view/NodeForm.js
```diff
@@ -421,10 +421,13 @@
     };
 
     var formIsInvalid = function () {
+      //  TODO remove this variable after Nodes and Junctions are imported.
+      var saveBtnDisabled = (Environment.name() === 'integration' || Environment.name() === 'production');
+
```
peltsi212: Disabling should be done only for production

PR #800, opened 2020-02-24
VIITE-2296 Prevent user discontinuity updates in case the links are not calculated

Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/194/consoleFull

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -1938,22 +1938,18 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       handleRoadNames(roadwayChanges)
       handleTerminatedRoadwayChanges(roadwayChanges)
       ProjectLinkNameDAO.removeByProject(projectID)
-      val projectLinksSet = projectLinks.map(_.roadNumber).toSet
       nodesAndJunctionsService.calculateNodePointsForProject(projectID, username = project.createdBy)
-      projectLinksSet
+      projectLinks.map(_.roadNumber).toSet
     } catch {
-      case e: ProjectValidationException => {
+      case e: ProjectValidationException =>
         logger.error("Failed to validate project message:" + e.getMessage)
         Set.empty[Long]
-      }
-      case f: SQLException => {
+      case f: SQLException =>
         logger.error("Failed to update roadways and linear locations with project links due to SQL error.", f)
         Set.empty[Long] // TODO Should we throw this exception so that the caller knows that something went wrong?
-      }
-      case ex: Exception => {
+      case ex: Exception =>
         logger.error("Failed to update roadways and linear locations with project links.", ex)
         throw ex
-      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1939,7 +1940,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       ProjectLinkNameDAO.removeByProject(projectID)
       val projectLinksSet = projectLinks.map(_.roadNumber).toSet
       nodesAndJunctionsService.calculateNodePointsForProject(projectID, username = project.createdBy)
-      return projectLinksSet
+      projectLinksSet
```
andre-policarpo: Change this to:
```
nodesAndJunctionsService.calculateNodePointsForProject(projectID, username = project.createdBy)
projectLinks.map(_.roadNumber).toSet
```

peltsi212: What is the reason to put projectLinks.map(_.roadNumber).toSet as last statement?
nodesAndJunctionsService.calculateNodePointsForProject works only when project changes are all done.
Is there change for project in this last statement?

More specifically project should change roadway_changes table and nodesAndJunctionsService.calculateNodePointsForProject depends on the fact that roadway_changes is populated.

andre-policarpo: In Scala, last statement always works as a return.
If we want the keep the method to return the same thing it was returning before, we should keep it.

This _projectLinksSet_ variable is not being used anywhere but the return, which means you can just remove that variable and just return its value.
```
val projectLinksSet = projectLinks.map(_.roadNumber).toSet
nodesAndJunctionsService.calculateNodePointsForProject(projectID, username = project.createdBy)
projectLinksSet```

Like I said in the comment:

```
nodesAndJunctionsService.calculateNodePointsForProject(projectID, username = project.createdBy)
projectLinks.map(_.roadNumber).toSet // this is being returned.
```

peltsi212: That scala behaviour is fine. I am just asking side effects of the last statement. There should not be any side effects after call to nodesAndJunctionsService.calculateNodePointsForProject.

andre-policarpo: > That scala behaviour is fine. I am just asking side effects of the last statement. There should not be any side effects after call to nodesAndJunctionsService.calculateNodePointsForProject.

In my honest opinion, its just to prevent one useless variable. It makes the code a little bit more clean.

Campanha: projectLinks.map(_.roadNumber).toSet is last statement because the method needs to return the roadNumbers in the changes to ultimately check the road network for the new added changed roadNumbers in RoadNetworkChecker

Campanha: > That scala behaviour is fine. I am just asking side effects of the last statement. There should not be any side effects after call to nodesAndJunctionsService.calculateNodePointsForProject.

projectLinks.map(_.roadNumber).toSet is last statement because the method needs to return the roadNumbers in the changes to ultimately check the road network for the new added changed roadNumbers in RoadNetworkChecker

PR #803, opened 2020-02-27
VIITE-2299 Added fix for dual roadway points handling when original roadwayNumbe…

Campanha:
…r is splitted


Campanha:
This commit i also done in 2303 branch. And there is also at least one test for it in that same branch

Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/195/console

PR #804, opened 2020-02-27
VIITE-2297 Refresh from master

skosonen:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/service/RoadLinkServiceSpec.scala
```diff
@@ -90,7 +88,7 @@ class RoadLinkServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
     }
   }
 
-  test("Test getCurrentAndComplementaryRoadLinksByMunicipality() When asking for info for a specific municipality Then return full municipality info, that includes both complementary and ordinary geometries") {
+  ignore("Test getCurrentAndComplementaryRoadLinksByMunicipality() When asking for info for a specific municipality Then return full municipality info, that includes both complementary and ordinary geometries") {
```
Campanha: should we ignore tests since we will still use complementary links from vvh?

PR #806, opened 2020-03-02
Handling case where VVH response is null.

skosonen:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/vvh/VVHClient.scala
```diff
@@ -366,10 +366,13 @@ trait VVHClientOperations {
       } catch {
         case _: IOException => Right(VVHError(Map(("VVH FETCH failure", "IO Exception during VVH fetch. Check connection to VVH")), url))
       } finally {
-        if (response != null)
+        if (response != null) {
           response.close()
-        if (response.getStatusLine.getStatusCode >= 300) {
-          return Right(VVHError(Map(("VVH FETCH failure", "VVH response code was <300 (unsuccessful)")), url))
+          if (response.getStatusLine.getStatusCode >= 300) {
```
Campanha: is it possible for the getStatusLine to be null?

andre-policarpo: I'll try to run import based on this branch. I think that workaround should prevent this null pointer from being thrown. Let's see..

skosonen: At least in this case it seems to be the response that is null:
"VVH FETCH failure -> VVH response was null."

PR #809, opened 2020-03-03
VIITE-2292_LINKID_and_MTKID_ search_to_ Search_box

peltsi212:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -223,6 +225,57 @@ class RoadAddressService(roadLinkService: RoadLinkService, roadwayDAO: RoadwayDA
...
+    }
+    return resultSeq
+  }
+
+  def locationInputParser(searchStringOption: Option[String]): Map[String, Seq[Long]] = {
+    val searchString = searchStringOption.getOrElse("")
+    val numRegex = """(\d+)""".r
+    val nums = numRegex.findAllIn(searchString).map(_.toLong).toSeq
+    var searchType = ""
+    if (nums.size == 0) {
```
Campanha: can transform var into val searchType = if.. else..

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -223,6 +225,57 @@ class RoadAddressService(roadLinkService: RoadLinkService, roadwayDAO: RoadwayDA
...
+    var resultSeq: Seq[Map[String, Seq[Any]]] = Seq.empty[Map[String, Seq[Any]]]
+    if (searchType == "road") {
+      val nums = parsedInput.head._2
+      if (nums.size == 3) {
+        searchResult = getRoadAddress(nums(0), nums(1), nums(2), None).sortBy(address => (address.roadPartNumber, address.startAddrMValue))
+        resultSeq = collectResult("road", searchResult, resultSeq)
+      } else if (nums.size == 2) {
+        searchResult = getRoadAddressWithRoadNumberParts(nums(0), Set(nums(1)), Set(Track.Combined, Track.LeftSide, Track.RightSide)).sortBy(address => (address.roadPartNumber, address.startAddrMValue))
+        resultSeq = collectResult("road", searchResult, resultSeq)
+      } else if (nums.size == 1) {
```
Campanha: should we deal with the else case option where num.size can be bigger than 3?

peltsi212: bigger values are now handled by returning resultSeq with inital value empty Sequence  as search result

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
peltsi212: Done

PR #811, opened 2020-03-04
VIITE-2327 Reviewed left joins and removed most of them.

skosonen:
Added some roadway.end_date is null and roadway.valid_to is null -restrictions.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodePointDAO.scala
```diff
@@ -97,7 +97,7 @@ class NodePointDAO extends BaseDAO {
                              FROM NODE_POINT NP
                              JOIN ROADWAY_POINT RP ON (RP.ID = ROADWAY_POINT_ID)
                              LEFT OUTER JOIN NODE N ON (N.NODE_NUMBER = np.NODE_NUMBER AND N.VALID_TO IS NULL AND N.END_DATE IS NULL)
-                             JOIN ROADWAY RW on (RW.ROADWAY_NUMBER = RP.ROADWAY_NUMBER) """
+                             JOIN ROADWAY RW on (RW.ROADWAY_NUMBER = RP.ROADWAY_NUMBER AND RW.end_date is NULL AND RW.VALID_TO IS NULL) """
```
Campanha: if we do this then we will not be able to get node points to terminated since the roadway(s) is/are already terminated

andre-policarpo: In my opinion: fetchByRoadwayPointIds, it's the only query where we need terminated roadways to be included. Others might work with end_date and valid_to validations.

skosonen:
Tests pass on Jenkins: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/200/console

skosonen:
Tests pass: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/201/console

PR #814, opened 2020-03-06
Viite 2299 roadway without modifications should not be splitted

Campanha:
completely unchanged single roadway with Unchanged + Transfer links should not create 2 new roadways for same roadway number


Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/208/console

PR #815, opened 2020-03-09
VIITE-2311 hotfix to expiration of node points when road type changes

andre-policarpo:
[Tests **OK**] http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/211/consoleFull


Campanha:
Seems good. But we should add a test fir this fix to prevent this case to come back in future.

andre-policarpo:
> Seems good. But we should add a test fir this fix to prevent this case to come back in future.

Agreed. I'll work on it today.

andre-policarpo:
[Tests **OK**: after changes] http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/213/consoleFull

PR #820, opened 2020-03-17
Viite 2317 junction numbers validations

andre-policarpo:
Includes: VIITE-2317, VIITE-2316, VIITE-2323
[Tests **Ok**] http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/219/consoleFull


viite-UI/src/model/SelectedNodesAndJunctions.js
```diff
@@ -1,6 +1,7 @@
 (function (root) {
   root.SelectedNodesAndJunctions = function (nodeCollection) {
     var current = {};
+    var junctionNumbersVerified = true;
```
skosonen: Should this be false initially?

andre-policarpo: Hmm, didn't thought about it. I would say that every nodes in the database should have their junctions with valid junction numbers. So, if we open a node its junctions should be 'verified'.

I actually don't like that local variable there, but it was the only way i found out to prevent the for each to run every time I need to check if junctions are valid or not.

skosonen: ok

viite-UI/src/view/NodeForm.js
```diff
@@ -140,63 +140,51 @@
...
+      var headers = function(options) {
         return '<tr class="node-junctions-table-header-border-bottom">' +
-          checkbox +
-          '<th class="node-junctions-table-header">NRO</th>' +
-          '<th class="node-junctions-table-header">TIE</th>' +
-          '<th class="node-junctions-table-header">AJR</th>' +
-          '<th class="node-junctions-table-header">OSA</th>' +
-          '<th class="node-junctions-table-header">ET</th>' +
-          '<th class="node-junctions-table-header">EJ</th>' +
+          ((options && options.checkbox) ? '<th class="node-junctions-table-header">' +
```
peltsi212: You might use backtick Template literals (Template strings) here instead of quote character. It works similar as scala s""" It is supported by chrome and Firefox and not by IE. We are not supporting IE so it can be used safely.

PR #827, opened 2020-03-23
VIITE-2345 Fixed three problems with calibration points.

skosonen:
- Updating of calibration point when we should expire old and create new
- One fetch without valid_to is null restriction
- One fetch that restricts also with roadway point id and cannot find old calibration point that is in different address


skosonen:
Tests pass on Jenkins: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/222/console

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/CalibrationPointsUtils.scala
```diff
@@ -118,4 +122,17 @@ object CalibrationPointsUtils {
     }
   }
 
+  def createCalibrationPointIfNeeded(rwPoint: Long, linkId: Long, calibrationPointLocation: CalibrationPointLocation,
+                                     calibrationPointType: CalibrationPointType, username: String): Unit = {
+    val existing = CalibrationPointDAO.fetch(linkId, calibrationPointLocation.value)
+    val (existingCorrect, existingDifferent) = existing.partition(cp => cp.roadwayPointId == rwPoint && cp.typeCode == calibrationPointType)
+    if (existingDifferent.nonEmpty) {
+      CalibrationPointDAO.expireById(existing.map(_.id))
```
skosonen: Found a bug here after merging: I should have expired "existingDifferent" and not all the "existing".

PR #828, opened 2020-03-24
Viite 2344 change calibration point type logic

andre-policarpo:
Tests **OK**: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/237/consoleFull

Should fix: VIITE-2329, VIITE-2338


Campanha:
@skosonen we dont want the junction point/roadway point addresses to affect addresses in roadways right? So "cp."TYPE" = ${CalibrationPointType.Mandatory.value}" should be enough?

skosonen:
Junction point calibration point: If there is already CP of type 2, we should not expire it and replace it with a new CP with type 1. (Now the createCalibrationPointIfNeeded would do this mistake) If there is an existing CP of higher type, we should keep that. If there is an existing CP having lower type, then that should be expired and replaced with CP type 1.

skosonen:
We want all calibration points to affect calculating of road addresses. We just need to implement the setting of junction point address so that we make junction point and calibration point use new roadway point. (We expire current junction points and calibration point and create new ones pointing to the corresponding roadway point.) This user given address must be between the calibration points before and after this junction point.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/LinearLocationDAO.scala
```diff
@@ -119,10 +120,10 @@ class LinearLocationDAO {
...
       select loc.id, loc.ROADWAY_NUMBER, loc.order_number, loc.link_id, loc.start_measure, loc.end_measure, loc.SIDE,
-      (SELECT RP.ADDR_M FROM CALIBRATION_POINT CP JOIN ROADWAY_POINT RP ON RP.ID = CP.ROADWAY_POINT_ID WHERE cp.LINK_ID = loc.LINK_ID AND loc.ROADWAY_NUMBER = rp.ROADWAY_NUMBER AND START_END = 0 AND cp.VALID_TO IS NULL) AS cal_start_addr_m,
-      (SELECT RP.ADDR_M FROM CALIBRATION_POINT CP JOIN ROADWAY_POINT RP ON RP.ID = CP.ROADWAY_POINT_ID WHERE cp.LINK_ID = loc.LINK_ID AND loc.ROADWAY_NUMBER = rp.ROADWAY_NUMBER AND START_END = 1 AND cp.VALID_TO IS NULL) AS cal_end_addr_m,
+      (SELECT RP.ADDR_M FROM CALIBRATION_POINT CP JOIN ROADWAY_POINT RP ON RP.ID = CP.ROADWAY_POINT_ID WHERE cp.LINK_ID = loc.LINK_ID AND loc.ROADWAY_NUMBER = rp.ROADWAY_NUMBER AND START_END = 0 AND cp.VALID_TO IS NULL AND cp."TYPE" = ${CalibrationPointType.Mandatory.value}) AS cal_start_addr_m,
+      (SELECT RP.ADDR_M FROM CALIBRATION_POINT CP JOIN ROADWAY_POINT RP ON RP.ID = CP.ROADWAY_POINT_ID WHERE cp.LINK_ID = loc.LINK_ID AND loc.ROADWAY_NUMBER = rp.ROADWAY_NUMBER AND START_END = 1 AND cp.VALID_TO IS NULL AND cp."TYPE" = ${CalibrationPointType.Mandatory.value}) AS cal_end_addr_m,
```
skosonen: I think these changes should be rolled back. According to my understanding we want all types of calibration points affect road address calculations.

skosonen:
> Junction point calibration point: If there is already CP of type 2, we should not expire it and replace it with a new CP with type 1. (Now the createCalibrationPointIfNeeded would do this mistake) If there is an existing CP of higher type, we should keep that. If there is an existing CP having lower type, then that should be expired and replaced with CP type 1.

This should be fixed now.

andre-policarpo:
Changed the way calibration points information are stored in `LinearLocation` object - currently, each `LinearLocation` doesn't only have the `addrM` information but also the `type` of the calibration point.

The best way I came up with was to create a new object `CalibrationPointReference` to store both `addrM` and `type`.

I kept the logic that each `LinearLocation` has a pair of calibration points, start and end.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/CalibrationPointsUtils.scala
```diff
@@ -125,10 +125,13 @@ object CalibrationPointsUtils {
...
-    val (existingCorrect, existingDifferent) = existing.partition(cp => cp.roadwayPointId == rwPoint && cp.typeCode == calibrationPointType)
-    if (existingDifferent.nonEmpty) {
-      CalibrationPointDAO.expireById(existing.map(_.id))
+
+    // The existing correct kind of calibration point has the same roadwayPointId and the same or a higher type.
+    val (existingCorrect, existingWrong) = existing.partition(cp => cp.roadwayPointId == rwPoint && cp.typeCode >= calibrationPointType)
+    if (existingWrong.nonEmpty) {
+      CalibrationPointDAO.expireById(existingWrong.map(_.id))
     }
+
```
andre-policarpo: You lost this >= in master changes I think.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/CalibrationPointsUtils.scala
```diff
@@ -135,4 +138,15 @@ object CalibrationPointsUtils {
     }
   }
 
+  def toCalibrationPointReference(cp: Option[ProjectLinkCalibrationPoint]): CalibrationPointReference = {
+    cp match {
+      case Some(x) =>
+        // TODO VIITE-2344 - CALIBRATION_POINT - this SOURCE should be verified
```
skosonen: What about this TODO-comment?

PR #830, opened 2020-03-27
Populate calibration point type 1 in DB

peltsi212:
VIITE-2351


digiroad2-oracle/src/main/resources/db/migration/V0_32__calibration_point_type_update.sql
```diff
@@ -0,0 +1,35 @@
+UPDATE CALIBRATION_POINT cpu SET "TYPE" = (SELECT CASE
+-- leave expired calibration points to original value 2
+		WHEN (cp.VALID_TO IS NOT NULL) THEN 2
```
andre-policarpo: I would remove this and just let the expired ones with the type they have - doesn't matter which is.
And add an `where` clause to the end.

peltsi212: fixed

digiroad2-oracle/src/main/resources/db/migration/V0_32__calibration_point_type_update.sql
```diff
@@ -0,0 +1,37 @@
+UPDATE CALIBRATION_POINT cpu SET "TYPE" = (SELECT CASE
+-- leave expired calibration points and user assigned points to original value
+-- [TYPE = 2] Includes templates, points where ADDR_M is equal to START_ADDR_M or END_ADDR_M of the road (road_number, road_part_number and track) and when ROAD_TYPE changes
+		WHEN (rp.ADDR_M = (SELECT MAX(roadAddr.END_ADDR_M) FROM ROADWAY roadAddr
```
skosonen: Why there is this MAX(roadAddr.END_ADDR_M)? I think all discontinuity places should be included.

peltsi212: fixed

PR #831, opened 2020-03-30
Viite 2347 wrong calibration points in project

Campanha:


Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/238/console

PR #834, opened 2020-04-01
Viite 2311 node point missing at roadtype change

Campanha:


Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/242/consoleFull

PR #837, opened 2020-04-02
VIITE-2288 JWT user authentication.

skosonen:
Assuming that the username comes in "x-amzn-oidc-identity" -header.


peltsi212:
I dot see any jwt functionality. Is there going to be also changes in frontend?

skosonen:
> I dot see any jwt functionality. Is there going to be also changes in frontend?

AWS LoadBalancer takes care of the authentication and then it adds JWT token in http headers. I assume that the username comes in the x-amzn-oidc-identity (subject) -header, so we don't even need to read the token itself but the username is all we need.

PR #838, opened 2020-04-02
VIITE-2286 Logging to stdout.

skosonen:


peltsi212:
Amazon and dockers in amazon support also file logging. What is the purpose of removing file logging?
For better compatibility in amazaon and docker logfile might be changed from 
logs/digiroad2.log to digiroad2.log
so that you do not have to create folder before logging

For readability I propose changing %level to %-5level
and %logger{10} to %logger{20}

skosonen:
> Amazon and dockers in amazon support also file logging. What is the purpose of removing file logging?
> For better compatibility in amazaon and docker logfile might be changed from
> logs/digiroad2.log to digiroad2.log
> so that you do not have to create folder before logging
> 
> For readability I propose changing %level to %-5level
> and %logger{10} to %logger{20}

Did the log readability changes. Not logging to file, since we run the Docker container in Fargate and at least now in the beginning we want to keep things simple and not write anything in disk.

PR #842, opened 2020-04-06
Viite 2365 tests rework after fix

Campanha:


Campanha:
all tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/265/console

PR #852, opened 2020-04-09
VIITE-2374 assign linear location id should be done in roadwayfiller …

Campanha:
…so the resulted project links could have the same linear location id


Campanha:
all tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/280/console

PR #857, opened 2020-04-16
VIITE-2377 created mismatching tracks exception

andre-policarpo:
[Tests **OK**] http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/286/consoleFull

- Current changes also fix **VIITE-2378**. Tested it and moved it to `code review` column in `jira`

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/TrackCalculatorStrategy.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/TrackCalculatorStrategy.scala
@@ -2,7 +2,7 @@ package fi.liikennevirasto.viite.process.strategy
 
 import fi.liikennevirasto.GeometryUtils
 import fi.liikennevirasto.digiroad2.asset.SideCode
-import fi.liikennevirasto.digiroad2.util.{MismatchingTracksException, RoadAddressException}
+import fi.liikennevirasto.digiroad2.util.{MissingTrackException, RoadAddressException}
 import fi.liikennevirasto.viite.NewIdValue
 import fi.liikennevirasto.viite.dao.ProjectCalibrationPointDAO.UserDefinedCalibrationPoint
 import fi.liikennevirasto.viite.dao.CalibrationPointSource.{ProjectLinkSource, RoadAddressSource, UnknownSource}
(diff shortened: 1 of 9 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/TrackCalculatorStrategy.scala
```diff
@@ -167,6 +167,8 @@ trait TrackCalculatorStrategy {
 
   protected def adjustTwoTracks(startAddress: Option[Long], leftProjectLinks: Seq[ProjectLink], rightProjectLinks: Seq[ProjectLink], calibrationPoints: Map[Long, UserDefinedCalibrationPoint],
                                 restLeftProjectLinks: Seq[ProjectLink] = Seq(), restRightProjectLinks: Seq[ProjectLink] = Seq()): TrackCalculatorResult = {
+    if (leftProjectLinks.isEmpty || rightProjectLinks.isEmpty)
```
skosonen, line 170: Should the exception be `MissingTrackException`? And the error message: "Missing left track." or "Missing right track"?

andre-policarpo: Since we never know which track is missing, that will imply writing unnecessary if/else or pattern matching statement. The more generic message were already being used, so i took it also for this case. I'm also logging the size of both tracks, in case it's needed for debugging in the future. We can always check which track is actually missing.

skosonen: In any case I think the exception should be `MissingTrackException`, because this is not about different roadway count on each track but other track completely missing.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/strategy/TrackCalculatorStrategy.scala
```diff
@@ -167,6 +167,8 @@ trait TrackCalculatorStrategy {
...
+      throw new MissingTrackException(s"Mismatching tracks, R ${rightProjectLinks.size}, L ${leftProjectLinks.size}")
```
skosonen, line 171: Logging message should be changed too: "Missing track. (R..., L...)".

PR #859, opened 2020-04-20
VIITE-2164 fix selecting node by project query

peltsi212:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -385,7 +386,39 @@ class NodeDAO extends BaseDAO {
...
+           ON N.NODE_NUMBER = NP.NODE_NUMBER
+         INNER JOIN ROADWAY_POINT RP
+           ON NP.ROADWAY_POINT_ID = RP.ID
+         INNER JOIN ROADWAY R
+           ON RP.ROADWAY_NUMBER = R.ROADWAY_NUMBER
+         INNER JOIN ROADWAY_CHANGES RC
+           ON (R.ROAD_NUMBER = RC.NEW_ROAD_NUMBER
+             AND R.ROAD_PART_NUMBER = RC.NEW_ROAD_PART_NUMBER)
+         WHERE RC.PROJECT_ID = $projectId
+         AND R.END_DATE IS NULL
```
Campanha, line 402: should we also add R.VALID_TO IS NULL ?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -366,7 +368,6 @@ class NodeDAO extends BaseDAO {
          AND R.END_DATE IS NULL
```
Campanha, line 368: should we also add R.VALID_TO IS NULL ?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -345,7 +345,9 @@ class NodeDAO extends BaseDAO {
   }
 
   // This query is designed to work in processing ROADWAY_CHANGES in phase where ROADWAY_CHANGES contains changes but other tables do not contain any updates yet
-  // First union part handles project changes and second part handles terminations
+  // First and third union part handles project changes and second and forth part handles terminations
+  // First and second union part finds nodes via junctions and third and fourth union find nodes via node_points
+  // We find also nodes that have end_date not null to support change detection for tierekisteri for expired nodes
```
skosonen, line 350: I guess the comment here should say "...for terminated nodes".

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
peltsi212: I removed R.VALID_TO IS NULL in commit 13.3.2002 I do not remember why I removed it. I put it back now. Lets hope it still works

PR #861, opened 2020-04-20
VIITE-2381 Letting the database set the valid_from value automatically

skosonen:
for the new rows of the node and road_name -tables.


skosonen:
Tests pass: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/296/console

PR #874, opened 2020-05-07
VIITE-2397 re assign roadwaynumbers for transfer cases in two tracks …

Campanha:
…if section have new length


Campanha:
tests passing at : http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/318/consoleFull

PR #875, opened 2020-05-07
Viite 2398 handle node point type

andre-policarpo:
Tests **OK** [here](http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/330/consoleFull).


andre-policarpo:
You can ignore 2393's changes - if you check them already - but I needed those to test/write this.
For code review: https://github.com/finnishtransportagency/viite/pull/875/commits/a27afa2cfbd2db513c894d9fb94a4178818c6f0f

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
```diff
@@ -496,90 +497,86 @@ class NodesAndJunctionsService(roadwayDAO: RoadwayDAO, roadwayPointDAO: RoadwayP
...
+
+      currentNodePoint match {
+        case Some(np) =>
+          val nodePointId = if (np.nodePointType == CalculatedNodePoint) Some(NewIdValue) else if (reversed) Some(np.id) else None
+          nodePointId.map { id =>
+            if (id == np.id) {
+              logger.info(s"Updated node point $id, due to reverse operation")
+              nodePointDAO.update(Seq(np.copy(beforeAfter = BeforeAfter.switch(np.beforeAfter))))
+            } else {
+              logger.info(s"Expired calculated node point ${np.id}")
```
peltsi212, line 524: I do not understand why CalculatedNodePoint is copied as RoadNodePoint. I think this should not be done.  Definition of CalculatedNodePoint can be found from VIITE-1862

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
```diff
@@ -496,90 +497,86 @@ class NodesAndJunctionsService(roadwayDAO: RoadwayDAO, roadwayPointDAO: RoadwayP
...
-                }
-              } else None
+            val headProjectLink = section.head
+            val headReversed = roadwayChanges.exists(ch => ch.changeInfo.target.startAddressM.nonEmpty && headProjectLink.startAddrMValue == ch.changeInfo.target.startAddressM.get && ch.changeInfo.reversed)
+
+            val headNodePoint = mappedRoadwayNumbers.find { rl =>
+              headProjectLink.startAddrMValue == rl.newStartAddr && headProjectLink.endAddrMValue == rl.newEndAddr && headProjectLink.roadwayNumber == rl.newRoadwayNumber
+            }.flatMap { rl =>
+              if (headReversed) {
+                nodePointDAO.fetchByRoadwayNumbers(Seq(rl.originalRoadwayNumber, headProjectLink.roadwayNumber))
```
peltsi212, line 552: nodePointDAO.fetchByRoadwayNumbers is used 4 times
I think nodePointDAO.fetchByRoadwayNumbers returns too many nodepoints
It returns calculated nodepoints and I think it should not do that. 
It uses left join to node which means that there is no need to node exist at all so it might return oprhaned nodepoints that are not connected to node.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
andre-policarpo: That was the point of this fix. If you check `VIITE-2398 Jiira task` you'll notice that yes, there was a Calculated Node Point that should be modified to be Road Node Point.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
andre-policarpo: nodePointDAO.fetchByRoadwayNumbers should return every node points (templates, calculated, road) so we know if we need to create a new one or not.

Basically if a node needs to be modified - due to:
1. Roadway point change
2. Before/After
3. Type

It's possible that the node doesn't exist yet, so it's a simple create, but if the node exists then we need to know it to expire it and create new one.

I tried my best to improve that code, before it was way much more raw code.
Although the query is written 4 times, it will only be called twice. One for the 1st point of the section, and another one for the last.

The main problem improving this code, is that in reverse cases, the address change and the beforeAfter is also reversed. Which makes it more complicated to found the right node.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
peltsi212: Calculated nodepoint is calculated as an average from junction m-values. I think Road node point is calculated with different algoritm. Because of that I think Road node point should be created as if there was no calculated node point. Now there is double logic for creating calculated node point. Original implementation takes care of calculated nodepoint both in project and in node tool. This new implementation does it only for project.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/NodesAndJunctionsService.scala
peltsi212: See other comment why calculated node point should not be fetched.

skosonen:
According to @peltsi212 the reason why the node point was not "updated" correctly was that the road address node point was not created. If the road address node point was created properly, then in the next step the process for creating the calculated node points would have cleared all calculated node points and then create any needed calculated node points. (In this case that process would have only deleted calculated node point and not create any new ones.)

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodePointDAO.scala
```diff
@@ -307,33 +324,6 @@ class NodePointDAO extends BaseDAO {
...
-          case Some(date) => dateFormatter.print(date)
-          case None => ""
-        })
-        ps.setLong(5, nodePoint.id)
-        ps.addBatch()
-    }
-    ps.executeBatch()
-    ps.close()
-    nodePoints.map(_.id).toSeq
-  }
```
andre-policarpo, line 335: Now it is created new row and expired old one, as it was suppose to.

PR #879, opened 2020-05-12
VIITE-2404 added Continuous check for end opposite link in the section

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/331/consoleFull

PR #880, opened 2020-05-12
Viite 2401 hotfix nodes and junctions queries

andre-policarpo:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodePointDAO.scala
```diff
@@ -95,7 +95,7 @@ class NodePointDAO extends BaseDAO {
                              FROM NODE_POINT NP
                              JOIN ROADWAY_POINT RP ON (RP.ID = ROADWAY_POINT_ID)
                              LEFT OUTER JOIN NODE N ON (N.NODE_NUMBER = np.NODE_NUMBER AND N.VALID_TO IS NULL AND N.END_DATE IS NULL)
```
Campanha, line 97: should we in the future remove the LEFT join from here?

skosonen: > should we in the future remove the LEFT join from here?

Have to test that. I remember there was at least one place where we purposefully left this "LEFT", because in that particular case it was correct and removing it would have broken things.

andre-policarpo: LEFT JOIN and LEFT OUTER JOIN are the same actually, so that can be changed to LEFT JOIN.
In this case, we can't change the LEFT JOIN, because it's needed. We're fetching both templates and not templates.

If we changed from LEFT JOIN to JOIN then we end up fetching only node points that have valid node_number, templates will not be fetched.

PR #887, opened 2020-05-18
Viite 2412 wrong starting point in new track change STAGING

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/345/console

PR #891, opened 2020-05-19
Viite 2394 missing wrong discontinuity code for parallel link

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/347/console

PR #893, opened 2020-05-21
VIITE-2415 merge continuous links with parallel Links. The inverse is…

Campanha:
… not true.


Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/349/consoleFull

PR #899, opened 2020-05-27
VIITE-2357 fix upgrade javascript libraries change uglify to terser. …

peltsi212:
…Delete integration tests


Campanha:
why should we delete integration-tests task?

peltsi212:
integration-tests task contains mainly commented out code. integration-tests task uses phantomjs and requirejs so that it was locking us to ecmascript 5

PR #900, opened 2020-05-27
Viite 2420 fix for node form save button

andre-policarpo:


viite-UI/src/less/core/forms.less
```diff
@@ -473,14 +487,15 @@ form.node .control-label-small {
...
 
-input.junction-number-input {
-  text-align: center;
-  width: 20px !important;
-  margin-right: 11px;
-  &:invalid, &:invalid:focus {
-    border-color: @error;
-  }
-}
+//input.junction-number-input {
```
Campanha, line 490: commented code should be deleted

peltsi212:
I think word hotfix should be reserved for updates to production that are done outside of normal publishing cycle. This save button problem has not been in production so this is not hotfix.

andre-policarpo:
> I think word hotfix should be reserved for updates to production that are done outside of normal publishing cycle. This save button problem has not been in production so this is not hotfix.

I actually didn't know how/when the word `hotfix` should be used. Thanks for the knowledge. I'll keep that in mind for future fixes and hotfixes.

viite-UI/src/less/core/forms.less
andre-policarpo: Done

PR #902, opened 2020-05-27
Viite 2416 wrong starting point for left track

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/363/console

PR #905, opened 2020-06-01
Viite 2421 could not find any nearest road address warning message

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/365/console

PR #910, opened 2020-06-03
Viite 2432 wrong starting point in combined road

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/369/consoleFull

PR #911, opened 2020-06-04
VIITE-2431 Parse username from JWT.

skosonen:


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/AdminApi.scala
```diff
@@ -70,6 +68,20 @@ class AdminApi(val dataImporter: DataImporter, implicit val swagger: Swagger) ex
...
+        Ok("Importing road names successful.\n")
+      } catch {
+        case e: Exception => {
+          logger.error("Importing road names failed.", e)
+          InternalServerError(s"Importing road names failed: ${e.getMessage}")
+        }
+      }
+    }
+  }
+
```
skosonen, line 84: Added also import_road_names task. (Not related to VIITE-2431)

PR #912, opened 2020-06-05
Viite 2440 missing junction template

Campanha:


Campanha:
tests passing at http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/371/consoleFull

PR #913, opened 2020-06-08
VIITE-2442 Remove node type 11 Pisaraliittymä

peltsi212:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -71,12 +71,6 @@ object NodeType {
     def displayValue = "Moniajoratainen liittymä"
   }
 
-  case object DropIntersection extends NodeType {
```
skosonen, line 74: No need to remove from here

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/NodeDAO.scala
```diff
@@ -23,7 +23,7 @@ sealed trait NodeType {
 
 object NodeType {
   val values: Set[NodeType] = Set(NormalIntersection, Roundabout, YIntersection, Interchange, RoadBoundary, ELYBorder, MultitrackIntersection,
-    DropIntersection, AccessRoad, EndOfRoad, Bridge, MaintenanceOpening, PrivateRoad, StaggeredIntersection, UnknownNodeType)
```
skosonen, line 26: No need to remove from here. Could you add also types 2, 6 and 9 in this list?

PR #914, opened 2020-06-08
VIITE-2446 Context path

skosonen:


skosonen:
Improved also logging of authenticated user.

peltsi212:
I think also Frontend needs to be updated gruntfile and backend-utils
What is the effect for integrations?

skosonen:
> I think also Frontend needs to be updated gruntfile and backend-utils
> What is the effect for integrations?

Frontend seems to work without any changes to gruntfile or backend-utils: https://apps.testivaylapilvi.fi/viite-dev/

Others are not yet integrated to AWS Viite and anyway those integrations will not be directly to Viite.

PR #924, opened 2020-06-16
VIITE-2454 Converted some date fields to timestamp fields.

skosonen:
Fixed some bugs when saving null dates.


skosonen:
Initial import worked and links are drawn on map.

PR #928, opened 2020-06-23
VIITE-2429_cache_breaker_configuration

peltsi212:


gruntfile.js
```diff
@@ -56,27 +56,30 @@ module.exports = function(grunt) {
       }
     },
     cachebreaker: {
-      css: {
+      dev: {
```
skosonen, line 59: Why the name "dev"? Why not something generic like "static"?

PR #933, opened 2020-06-30
VIITE-2463 terminated links should be adjusted for change table

Campanha:


Campanha:
tests passing here: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/383/console

PR #934, opened 2020-07-01
VIITE-2467 Changed exception handler

Campanha:


Campanha:
tests passing at: http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/384/console

PR #936, opened 2020-07-08
VIITE-2425: ignore calibration point recalculation for history road addresses if they are terminated

oskarmalmstrom:


Campanha:
Dunno why the Review Button doesnt appear. But this seems fine.

PR #937, opened 2020-07-08
Viite 2125 new vkm interface in use

oskarmalmstrom:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ViiteVkmClient.scala
```diff
@@ -42,7 +71,7 @@ class ViiteVkmClient {
     val url = new URL(getRestEndPoint)
     if (url.getHost == "localhost") {
       // allow ssh port forward for developing
-      post.setHeader("Host", "oag.liikennevirasto.fi")
+      post.setHeader("Host", "testijulkinen.vayla.fi")
```
peltsi212, line 74: Probably this should not be changed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ViiteVkmClient.scala
```diff
@@ -28,6 +30,33 @@ class ViiteVkmClient {
 
   private val client = HttpClientBuilder.create().build
 
+  def get(path: String, params: Map[String, String]): Either[Any, VKMError] = {
+
+    val url =
+      if(params.nonEmpty)
+        getRestEndPoint + path + "?"+params.map(p => if (p._2.nonEmpty) p._1 + "="+ p._2).mkString("&")
```
peltsi212, line 37: Can you use URIBuilder to build url
See for example https://stackoverflow.com/questions/9907161/commons-httpclient-adding-query-string-parameters-to-get-post-request

PR #940, opened 2020-07-09
VIITE-2482 Got rid of old CI and Oracle specific stuff. Fixed fixture reset.

skosonen:
Properties are now given in /conf/env.properties in local development and in environment variables in other environments.
Fixed also showing of version information.


skosonen:
Working on getting the fixture reset work. We are using so old version of Flyway that it doesn't recognize Postgis extension but tries to remove it as well on cleanup.

skosonen:
All tests pass. (Couple of tests are ignored though.)

PR #942, opened 2020-07-13
Viite 2470 node tool performance

oskarmalmstrom:


viite-UI/src/view/NodeLayer.js
```diff
@@ -569,7 +569,6 @@
       } else if (layer === 'node') {
         setGeneralOpacity(1);
         addInteractions();
-        showLayer();
```
oskarmalmstrom, line 572: showLayer is already called in MapView.js line 107

viite-UI/src/view/NodeForm.js
```diff
@@ -409,8 +409,8 @@
     };
 
     var closeNode = function (cancel) {
+      eventbus.off('change:nodeName change:nodeTypeDropdown change:nodeStartDate junction:validate junction:setCustomValidity');
```
peltsi212, line 412: This list contains now more events than before. I think eventbus.off should be called for all places where there is eventbus.on for example node:selected is still missing

oskarmalmstrom: The listener for node:selected is only created when `bindEvents()` is called on initialization so there is no need to remove the listener. But for the listeners created inside node:selected it's true, I will add them.

oskarmalmstrom: ...but you are right, the node:selected listener still needs to be removed. I will correct that.

PR #952, opened 2020-07-22
VIITE-2484 fix eslint warnings

peltsi212:


viite-UI/src/map/CalibrationPointMarker.js
```diff
@@ -2,16 +2,12 @@
   root.CalibrationPoint = function(data) {
     var cachedMarker = null;
     var cachedDirectionArrow = null;
-    var defaultMarkerGraphics = {
-      externalGraphic: "images/calibration-point.svg"
-    };
     var firstCustomCalibrationPointValue=4;
     var createCalibrationPointMarker = function () {
+      const marker = new ol.Feature({
```
oskarmalmstrom, line 7: Is it OK to use ES6 features? Does this work with IE11?

peltsi212: Viite does not support IE11. It is ok to use es6 and newer features

skosonen:
This should be merged only after VIITE-2479 has been taken to production.

PR #953, opened 2020-07-23
Viite 2125 reapply new vkm interface

oskarmalmstrom:


oskarmalmstrom:
This feature is dependent on the launch of the new VKM interface to production

PR #957, opened 2020-08-04
VIITE-2506 Fix calibration point types

skosonen:


skosonen:
All tests pass on Jenkins #409.

PR #963, opened 2020-08-25
viite-2424 validate new projectlink

peltsi212:
Test covers case where junction is missing in connection point.
There is no test for case where there is junction in connection point.


skosonen (approved):

skosonen:
Running tests...

skosonen:
Two tests failed.

PR #968, opened 2020-08-28
VIITE-2513 Update junction point address

skosonen:


andre-policarpo (approved):

andre-policarpo:
Looks good to me. Although validations about the addresses should go up in prod by the same time that this one.

PR #970, opened 2020-09-09
VIITE-2521

andre-policarpo:
**Update**: Allowed new links to be added in two track roads after split implementation.

Tests **Ok** [here](http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/421/consoleFull).

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -1459,7 +1459,18 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       }
     }
 
-    val projectLinks = projectLinkDAO.fetchProjectLinks(projectId)
+    val (originalProjectLinks: Seq[ProjectLink], splitLinks: Seq[ProjectLink]) = projectLinkDAO.fetchProjectLinks(projectId).partition(_.connectedLinkId.isEmpty)
+
+    val projectLinks = originalProjectLinks.map { pl: ProjectLink =>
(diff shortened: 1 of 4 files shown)


AnttiJokela (approved):

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1499,6 +1510,8 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
       val (generatedLinks, adjustedLinks) = (recalculated ++ assignedTerminatedRoadwayNumbers).partition(_.id == NewIdValue)
 
+      roadwayChangesDAO.clearRoadChangeTable(projectId)
```
AnttiJokela, line 1513: I'm not sure if I understood the need for this, and why it was not needed before?

andre-policarpo: **CHANGE_TABLE** data, for a certain project, is produced from scratch every single time an user changes anything within project links. Because, in this process we need to revert all the project links that were split we will need to delete some lines from the table **PROJECT_LINK**.

Because change_table and project_link are connected by **FK** reference, the best way to do it, is to clear the change table first:

`roadwayChangesDAO.clearRoadChangeTable(projectId)
projectLinkDAO.removeProjectLinksById(splitLinks.map(_.id).toSet)`

AnttiJokela: Ok, got it.

PR #977, opened 2020-10-12
VIITE-2533 Nodes are missing from TR

AnttiJokela:
When updating Node on a road part that has both single and dual tracks, viite incorrectly places the node point on a single track roadway (results are ordered by track ASC). In some cases the point address then becomes invalid because the point M value does not belong to the possible roadway M values. TR does not find an address for the track 0 and therefore ignores the node point.

Added validation of the M value to be incuded in the roadway that the node point is going to be added.


LauriLitmanenSW (approved):

AnttiJokela:
Test OK in Jenkins

PR #978, opened 2020-10-13
VIITE-2533 Node point on roadway without junction

AnttiJokela:
Handle a special case when calculated nodepoint resides on a roadway that does not have a junction on it.


ao-sitowise (commented):
Ok.

AnttiJokela:
Tests OK

PR #990, opened 2021-01-20
Dockerfile update

ao-sitowise:
A fix to Docker build to use Alpine version 3.12.3 instead of latest.


AnttiJokela:
Should fix all the usages of alpine image.

AnttiJokela (approved):

PR #992, opened 2021-02-01
Viite-2564 Extra Nodepoint

LauriLitmanenSW:
headReversed flag now also checks endAddrM of the projectlink when determining if the link has been reversed. Earlier version only checked the startAddrM which was = 0. This resulted the flag to be set true when in reality it should've been false. The code checked another links startMValue and when it was also 0, it thought it had found the project link in question and set the flag to be True because the wrong project link had been reversed. This fixed issue Viite-2564 where an extra nodepoint was created due to Viite not finding the old one because of the headReversed flag being at incorrect state when fetching the nodepoint.


LauriLitmanenSW:
Tests OK [here](http://livibuild04.vally.local/jenkins/job/viite/job/Run-Tests/494/consoleFull)

AnttiJokela (approved):

PR #1006, opened 2021-03-02
Refactoring, roadlinkbox text from road type to administrative class,…

LauriLitmanenSW:
… ViiteApi to map road type to administrativeClassId instead of roadTypeId, changed roadTypeId occurrences to administrativeClassId, few other occurrences where road type needed to be changed to administrative class.


marjukka-kokkonen-sitowise (commented):
Muutokset ok

PR #1017, opened 2021-03-18
Postgis ao build test

ao-sitowise:
Docker tags and paths need to be update for aws-postgis deployment in Buildspec.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -574,7 +574,11 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
                 Map(discontinuityAtEnd._1 -> discontinuityAtEnd._2)
             } else
               Map()
+
+            projectLinkDAO.fetchProjectLinks(projectId).map(pl => pl.toString).foreach(logger.debug)
```
AnttiJokela: remove these

PR #1020, opened 2021-03-29
A test fix.

ao-sitowise:
The test "Test changeDirection() When projectLinks are reversed the track codes must switch and start_addr_m and end_addr_m should be the same for the first and last links" fails some times.
This fixes ordering after db fetch to match ordering of projectLinksBefore and projectLinksAfter.


AnttiJokela (approved):

ao-sitowise:
VIITE-2575 fix sorting condition.

ao-sitowise:
VIITE-2575 fix sorting condition.

PR #1051, opened 2021-07-06
Viite-2589 add geometry length and roadaddress length to link property form

LauriLitmanenSW:
added geometry length to link property and project edit forms, added addr length to link property form


marjukka-kokkonen-sitowise (approved):
Checked & tested; looking good!

PR #1066, opened 2021-08-31
Viite-2606 use https connection to OAG instead of http

LauriLitmanenSW:
use https connection to OAG instead of http


AnttiJokela (approved):

marjukka-kokkonen-sitowise (approved):
taskRoleArn updated, already running in AWS, all is good, may be merged

PR #1072, opened 2021-09-03
Fix buildspec cluster to point to new ALB stack Cluster (fix case error)

marjukka-kokkonen-sitowise:


sitowisematti:
seems ok

sitowisematti (commented):
ok

PR #1083, opened 2021-09-16
Viite-2631 replace normal validation with high priority validation when reverting changes to projectlink(s)

LauriLitmanenSW:
replace normal validation with high priority validation when reverting changes to projectlink(s)


LauriLitmanenSW:
Tests OK

marjukka-kokkonen-sitowise (approved):

PR #1088, opened 2021-09-20
Viite-2626 refresh view after recalculation

LauriLitmanenSW:
refresh view when recalculate button is clicked  (after backend is done recalculating.)


LauriLitmanenSW:
Tests OK

AnttiJokela (approved):

PR #1124, opened 2021-11-03
VIITE-2552 junction and node point expiration bug fix

LauriLitmanenSW:
Terminated roadway was given a new roadway number (not the same roadway number that project link had) and
roadway points were updated with the project links roadway number.

 That caused a bug where obsolete node and junction points weren't terminated because Viite couldn't find the roadway points that the nodes and junction points were on.


LauriLitmanenSW:
Tests OK

ao-sitowise (approved):

PR #1130, opened 2021-11-11
VIITE-2588 single click selects whole road part…

LauriLitmanenSW:
…and not just the homogenic part of the road part

Single click selects whole road part and not just the homogenic section of the road part

link property form shows all the unique link property values on the selected road part if more than one link is selected (e.g. the whole road part with multiple track codes).

refactored dynamic field


marjukka-kokkonen-sitowise (approved):
👍

PR #1132, opened 2021-11-16
VIITE-2588 fix link group partitioning

LauriLitmanenSW:
Fixed a bug where road link duplicates appeared in roadCollection.js roadlinkGroups.

Link groups were partitioned by homogeneous section and if track wasn't same on two different road links then they were put in different groups.
Now that single click selects whole road part regardless of different tracks, we dont have to partition by track anymore.


LauriLitmanenSW:
Tests OK

ao-sitowise (approved):
same -> the same... ;)

ao-sitowise (approved):

PR #1136, opened 2021-11-24
task-definition script related cosmetics

marjukka-kokkonen-sitowise:
Removed unnecessary lines from task-def script files. Renaming dev-task-def to contain dev in the file name. Corrected QA task-def-path in the script.


ao-sitowise (approved):
ok

PR #1143, opened 2021-12-20
VIITE-2664 bug fixes

LauriLitmanenSW:
Fixed a bug where if the last selected link was deselected, all links would stay pale instead of returning to deep colored.
Fixed a bug where selected links did not stay as selected when the map was moved and more links were selected.

Grunt tests OK


sitowisematti (approved):
ok

PR #1171, opened 2022-02-15
VIITE-2694 fix changeTable open bug

LauriLitmanenSW:
Backend: changeTable is now recalculated for every projectState except Accepted, InUpdateQueue and UpdatingToRoadNetwork

UI: add check to a condition


LauriLitmanenSW:
tests OK

ao-sitowise (approved):

PR #1356, opened 2022-11-29
VIITE-2374 fix renamed method calls

SitowiseValtteriV:


ao-sitowise (approved):
Needs RoadAddressService update

PR #1385, opened 2023-01-10
VIITE-2929 fix validation errors missing from UI bug

LauriLitmanenSW:


marjukka-kokkonen-sitowise (approved):
The changes itself ok.
Some thoughts about combining setProjectErrors, and setAndWriteProjectErrorsToUser, but those are of lower priority.

PR #1459, opened 2023-03-27
VIITE-2996 fix date validation

LauriLitmanenSW:
z-index fix so that the overlays are the top most elements on the screen when displayed

added space to string

### DIFF
--- a/viite-UI/src/utils/date-utils.js
+++ b/viite-UI/src/utils/date-utils.js
@@ -143,7 +143,7 @@
   };
 
   dateUtils.isDateInYearRange = function (date, minYear, maxYear) {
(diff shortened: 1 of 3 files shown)


viite-UI/src/utils/date-utils.js
```diff
@@ -143,7 +143,7 @@
   };
 
   dateUtils.isDateInYearRange = function (date, minYear, maxYear) {
-    return date.getFullYear() > minYear && date.getFullYear() < maxYear;
+    return date.getFullYear() > (minYear - 1) && date.getFullYear() < (maxYear + 1);
```
marjukka-kokkonen-sitowise, line 146: hmm...
 >= / <= may be clearer than +-1?

marjukka-kokkonen-sitowise (approved):

PR #1473, opened 2023-04-13
VIITE-3001 date input fields now support cursor moving with  arrow ke…

LauriLitmanenSW:
…ys and using tab to move to next elements.

deleted obsolete code


marjukka-kokkonen-sitowise (approved):
👌

PR #1501, opened 2023-05-12
VIITE-3034 fix build

LauriLitmanenSW:
add fixture reset test to qa buildspec file
rename dev buildspec file


marjukka-kokkonen-sitowise (approved):
👌

PR #1559, opened 2024-01-30
Viite 3084 date validation

jounijoh:
Added isFinnishDateString function to [date-utils.js](https://github.com/finnishtransportagency/viite/compare/VIITE-3084_date_validation?expand=1#diff-d34e60de5601adc53af074a8fc4faccc86b25d8c49f995c904a6ded7273d8097) to check for correct date format of dd.mm.yyyy with 31 days and 12 months. 

It's now used in validateDate function in [RoadAddressBrowserWindow.js](https://github.com/finnishtransportagency/viite/compare/VIITE-3084_date_validation?expand=1#diff-f40912904efc8cc9ef7b8f3dd1d271c404b7dea77d36e44a69818f8001e0b9cc) to check for correct format of the input string.  

Modified willPassValidations to pass input value as a string instead of date object to validateDate.

Added event listener to clear the date error message when user starts typing the date again.

### DIFF
--- a/viite-UI/src/utils/date-utils.js
+++ b/viite-UI/src/utils/date-utils.js
@@ -18,6 +18,12 @@
     return s ? moment(s, dateUtils.FINNISH_DATE_FORMAT) : null;
   };
 
+  dateUtils.isFinnishDateString = function (dateString) {
+    // Regular expression to match date format with day 1-31 and month 1-12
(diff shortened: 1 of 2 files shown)


marjukka-kokkonen-sitowise (approved):
👌

PR #1637, opened 2024-07-17
Viite-3144 aws code artifact set up 

jounijoh:
- Added CloudFormation stack to create codeArtifact repositories.
- Modified build.scala to use codeArtifact as a source for packages conditionally when wanted.
- Added CodeArtifactSettings.scala to centralize all CodeArtifact related configurations.
- Packages are now conditionally downloaded based on the presence of the CodeArtifact token.
- Packages are now sourced from CodeArtifact when the token is set, otherwise from public repositories.
-  build.scala projects now use projectSettings, that are common settings for all the projects
- Updated buildspec to use CodeArtifact for npm package management.
- Configured buildspec to download SBT from S3 bucket instead of public repositories.
- Integrated CodeArtifact authentication into the build pipeline for both npm and Maven packages.


project/build.scala
```diff
@@ -45,21 +47,30 @@ object Digiroad2Build extends Build {
...
+  val geoToolsDependencies: Seq[ModuleID] = Seq(
+    "org.geotools" % "gt-graph" % GeoToolsVersion,
+    "org.geotools" % "gt-main" % GeoToolsVersion,
+    "org.geotools" % "gt-referencing" % GeoToolsVersion,
+    "org.geotools" % "gt-metadata" % GeoToolsVersion,
+    "org.geotools" % "gt-opengis" % GeoToolsIFVersion,
+    "jgridshift" % "jgridshift" % JgridshiftVersion
+  )
+
+  // Common settings for all projects
```
marjukka-kokkonen-sitowise, line 59: <3 😃

marjukka-kokkonen-sitowise (approved):

PR #1672, opened 2024-08-29
Viite 3184 aws notifications for cost anomalies

jounijoh:
Added cloudFormation template for cost anomaly detection
- Daily checks if costs go over 15% of the expected cost
- One notification per day for all daily anomalies that exceed the 15% threshold


marjukka-kokkonen-sitowise (approved):
👌

PR #1685, opened 2024-09-19
Viite-3192 Update AWS Cost Anomaly Detection Thresholds

jounijoh:
Changed the AWS Cost Anomaly Detection threshold to trigger alerts only when the daily cost of a service doubles and the cost exceeds 2 dollars over the average cost to avoid unnecessary alerts.


marjukka-kokkonen-sitowise (approved):
Jes, kaikkein pienimmät ilmot pois - vaikka sinänsä sympaattisia ovatkin :D

PR #1707, opened 2024-11-12
VIITE-3247 Fix node edit bug

LauriLitmanen:
Roles are now fetched before the initialization of NodeForm.js.


jounijoh (approved):
👍

marjukka-kokkonen-sitowise (approved):
👌

PR #1717, opened 2024-12-02
Viite 3260 fix to "move to junction/node template form bug/jam"

jounijoh:
moveToCollection:
- Refactored moveToLocation to use async/await for better flow control and error handling
- Combined node data handling from separate callbacks into single Promise-based flow to ensure reliable data processing sequence as it was causing issues 
- Fixed spinner management and added identifier 'moveToLocation' for it

NodeForm.js:
- modified cancel buttons of node and junction edit templates to clear the url to previous state
- removed applicationModel.removeSpinner(); as it is handled in moveToCollection solely now

NodeSearchForm.js: 
- Changed 'click', '.junction-template-link', function event  to use window.location.hash with preventDefault instead of eventbus.trigger to Prevent double event triggering from both click and URL change

### DIFF
--- a/viite-UI/src/model/NodeCollection.js
+++ b/viite-UI/src/model/NodeCollection.js
@@ -63,58 +63,88 @@
       });
     };
 
-    this.moveToLocation = function (template) {
-      if (!_.isUndefined(template)) {
-        applicationModel.addSpinner();
-        locationSearch.search(`${template.roadNumber} ${template.roadPartNumber} ${template.addrM} ${template.track}`).then(function (results) {
-          if (results.length >= 1) {
-            const result = results[0];
-            eventbus.trigger('coordinates:selected', {
-              lon: result.lon,
-              lat: result.lat,
-              zoom: zoomlevels.minZoomForJunctions
-            });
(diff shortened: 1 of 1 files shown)


viite-UI/src/model/NodeCollection.js
```diff
@@ -63,59 +63,89 @@
...
+            }
+            resolve(fetchedNodesAndJunctions);
+          });
+          // Trigger node data fetch for the selected location
+          eventbus.trigger('nodeLayer:fetch');
+        });
+      } catch (error) {
+        console.error('Operation failed:', error);
+      } finally {
+        // Ensure spinner is always removed
```
marjukka-kokkonen-sitowise, line 144: 👌

marjukka-kokkonen-sitowise (commented):
Mun silmään näyttää hyvältä.
👌 
Jonkun enemmän js-natiivin kannattanee silti lukea läpi myös, mun lukemisellani jää ymmärrykseen aukkoja.

LauriLitmanen (approved):

PR #1719, opened 2024-12-03
VIITE-3273 better JOIN clause for SQL query to prevent addr mismatche…

LauriLitmanen:
…s (negative lengths / wrong address M ranges)


marjukka-kokkonen-sitowise (approved):
👌

PR #1725, opened 2024-12-18
VIITE-3269 switch to CSV export (instead of SheetJS excel export.)

LauriLitmanen:
SheetJS library didn't behave as we expected, causing all kinds of problems with dates.

New implementation allows users to download data in a CSV-file. The data should stay untouched this way and there shouldn't be any issues with dates no more.


marjukka-kokkonen-sitowise (approved):
👌 
Eipä lisättävää.
Ei nillejäkään. :)

PR #1729, opened 2025-01-13
Viite 3094/viite 3308 scalike jdbc nested transactions fix

jounijoh:
Nested transactions fix
- Nested transactions are now prevented with SessionProvider in same way as they were with Slick implementation
- Modified comments to point out prevented nested transactions
- Added runWithFutureTransaction for future transactions
- Nested sessions are allowed for read only operations

Refactored userProvider session handling
- Renamed PostGISUserProvider and its Spec files to UserProviderDAO
- Deleted RequestHeaderAuthentication.scala as it is not used anywhere
- Session handling for userProvider db operations are now handled in Authentication instead of the DAO layer
- Removed unnecessary AutoCommit block from UserProviderDAO test case
- The test now uses the actual DAO-method for deletion instead of SQL query on the test it self


viite-backend/database/src/main/scala/fi/vaylavirasto/viite/postgis/SessionProvider.scala
```diff
@@ -21,21 +42,29 @@ object SessionProvider {
...
    * @param f         The block of code to execute.
    * @tparam T        The return type of `f`.
-   * @return          The result of executing `f`.
-    */
+   * @throws IllegalStateException if called within an existing transaction
+   */
   def withSession[T](dbSession: DBSession)(f: => T): T = {
-    val previousSession = threadLocalSession.get()
-    threadLocalSession.set(dbSession)
+    if (isTransactionOpen && !dbSession.isReadOnly) { // Allow nested transactions for read-only operations
```
marjukka-kokkonen-sitowise, line 54: p.o.

// Prevent nested transactions for write operations

marjukka-kokkonen-sitowise (approved):
👌

marjukka-kokkonen-sitowise (approved):
Vielä paremmalta vaikuttaa. :)

PR #1732, opened 2025-01-17
VIITE-3320  Now case class RoadwaySection uses AddrMRange.

marjukka-kokkonen-sitowise:
(Also: Moving RoadwaySection where it is used: ProjectDeltaCalculator.)


jounijoh (approved):
👌

PR #1739, opened 2025-01-24
VIITE-3094/VIITE-3330 data fixture scalike jdbc migration

jounijoh:
DataFixture migration:
- DataFixture now uses ScalikeJDBC
- runUpdateToDbSlick is now defined in DataImporter as it is the only part of the app that uses the Slick version
- Municipality ids are now fetched from MunicipalityDAO instead of queries.scala as they are not new sequences

NodeDAO cleanup:
- Removed unused table alias
- expireById return type changed to Int instead of Unit
- Removed unused fetchId method


marjukka-kokkonen-sitowise (approved):
Näyttää hyvältä, ei ees nillejä. 👌

PR #1751, opened 2025-02-05
VIITE-3346 nodes and junction wont update when map is moved fix to nextRelease

jounijoh:
Hot fix to nextRealease:

Nodes and Junctions now appear on the map correctly when the form is opened or map moved. 

- node:fetched event handler is added back to its own event listener to handle the fetched node data separately.
- The moveToLocation function now waits for the node:fetched event to be triggered and resolves the promise with the fetched data.
- This ensures that node:fetched triggers correctly also with fetchedNodes function to load all nodes and junctions, not just the selected ones
- Removed unnecessary spinnerEvent(fetching) as only closing the form happens after its triggered

### DIFF
--- a/viite-UI/src/model/NodeCollection.js
+++ b/viite-UI/src/model/NodeCollection.js
@@ -96,55 +96,62 @@
         });
 
         // Fetch node data for the selected location
-        await new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


marjukka-kokkonen-sitowise (approved):
👌

PR #1760, opened 2025-02-21
VIITE-3304/VIITE-3366: Removed Slick from Code Base

jounijoh:
- Removed Slick library and related ghost files from the code base


marjukka-kokkonen-sitowise (approved):
Joo pois vaan. :)

PR #1771, opened 2025-03-04
VIITE-3346: fix to load nodes and junctions to map correctly when map is moved (#1751)

marjukka-kokkonen-sitowise:


marjukka-kokkonen-sitowise:
Merge just before next postgis->NextRelease Merge

PR #1822, opened 2025-06-12
VIITE-3015 Fix run configurations

LauriLitmanen:
Switched from deprecated run-main to the correct runMain task key for compatibility with SBT 1.1.4


jounijoh (approved):
👍

PR #1835, opened 2025-07-10
VIITE-3531 Fix TooManyRowsException in fetchLatestId DAO function

LauriLitmanen:
When a Node was edited, both a history row and a new "current" row were created. As a result, fetching nodes by node number could return multiple rows. The runSelectSingleOption method uses .single(), which expects exactly one result, leading to a TooManyRowsException when multiple rows were returned.

This fix updates the logic to select only the latest node ID from the already ordered query result, ensuring a single result is returned.


LauriLitmanen:
Tests OK

PR #1863, opened 2025-08-26
VIITE-3541 unaddressed links are shown on the map by default. 

LauriLitmanen:
Added the "checked" attribute to the unaddressedRoadsVisibleCheckbox


Jackgito (approved):
OK

PR #1998, opened 2026-04-09
VIITE-3937 fixed sql bug related to dates (artifact from oracle age?)

eikrtt:
Fixed bug where /api/viite/integration/nodes_junctions/changes did not work correctly

Needs review and assurance that Viite works after this change.


JuhaniManninen (approved):
Testattu lokaalissa toimivaksi

PR #2022, opened 2026-05-26
Added a check to filter out cases where calculating average ending va…

PellinenErkko:
…lues for unrelated opposite tracks resulted in negative values.


JuhaniManninen (approved):
OK

PR #2026, opened 2026-06-02
VIITE-3958 Use EVK0 when municipality is not available from KGV

PellinenErkko:


PellinenErkko:
Changed a previous faulty default left over from the EVK0 bug hunt back to EVK0-default under circumstances. 
This particular case seems to have only arisen when road links from outside of Finlands borders were returned by the BoundingBox. They naturally don't have a finnish municipality code, and are thereby safe to be set to default into EVK0. 

Also added a logging to trigger when handling road links with municipality code <= 0 to clarify where the EVK0 value originates from.

JuhaniManninen (approved):
OK

PR #2037, opened 2026-06-15
Changes to terminated roadway handling.

PellinenErkko:
Changed the logic in RoadwayFiller.scala and ProjectService.scala to allow roadways and linear locations to be created for terminated sections.


JuhaniManninen (approved):
OK

PR #2059, opened 2026-08-19
VIITE-3936 Expanded the implementation.

PellinenErkko:
Expanded the previous implementation to allow similar cross-roadMaintainer fetching of data for other target types.


JuhaniManninen (approved):
OK

